PR #9, opened 2018-05-07
fix typo in associativity of addition inf. rule

Fingerzam:
I think this is a typo.


wadler:
Already spotted by someone else. Thank you!

PR #10, opened 2018-05-07
remove repeated text

Fingerzam:
It looks like these lines were repeated twice in a row.
Unfortunately the diff below is not the optimal one.


wadler:
Thank you! -- P

PR #20, opened 2018-05-14
Fix typo: products distributes -> product distributes

Fingerzam:


Fingerzam:
Should this actually be "Products distribute"?

PR #25, opened 2018-06-18
Fix operator in `∸-+-assoc` exercise

jsyeo:
The equation that shows monus associates with addition should be:

```
(m ∸ n) + p ≡ m ∸ (n + p)
```

it was previously

```
(m ∸ n) ∸ p
```

on the left hand side.


jsyeo:
Oops. my bad. This is correct. I threw my mathematics out of the door. LOL.

PR #26, opened 2018-07-05
[ Induction ] Fix typo

jonaprieto:
of of typo


jonaprieto:
On another hand, I was reading at the end of `+-comm` proof.

> Agda requires that identifiers are defined before they are used

Yes, but not always, remember the keyword `mutual` that allows us to write such
things like:

```agda
mutual
  theorem1 = byLemma

  byLemma = proof
```

PR #29, opened 2018-07-17
Update Relations.lagda

Kwezan:
to be consistent with previous definition and the description below i.e.: "Base case: for all naurals n, ....."


wadler:
Thank you!

PR #31, opened 2018-07-18
additional explanation on rewrite

Kwezan:
"Rewriting by +-comm m p and +-comm n p converts m + p ≤ n + p into p + m ≤ p + n, which is proved by invoking +-monoʳ-≤ p m n m≤n."
The above sentence from the chapter "Relation: Monotonicity" as a conclusion, is not easy to grasp until someone knows how the notation "rewrite" works or has previously been familiar with the material from the chapter "Equality: Rewriting expanded".
In addition, the definition of monotonicity using the notation "with" could serve as an additional exercise, for example:
+-monoˡ-≤1 : ∀ (m n p : ℕ) → m ≤ n → m + p ≤ n + p
+-monoˡ-≤1 m n p mn with    m + p |  +-comm m p  | n + p  |  +-comm n p
...                                      | .(p + m)  | refl | .(p + n) | refl = +-monoʳ-≤ p m n mn

### DIFF
--- a/src/plfa/Relations.lagda
+++ b/src/plfa/Relations.lagda
@@ -404,6 +404,8 @@ result and the commutativity of addition.
 \end{code}
 Rewriting by `+-comm m p` and `+-comm n p` converts `m + p ≤ n + p` into
 `p + m ≤ p + n`, which is proved by invoking `+-monoʳ-≤ p m n m≤n`.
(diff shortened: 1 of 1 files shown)


wadler:
Thanks. What I really should do is rewrite using chains of equalities, but that is tricky because here chains of _inequalities_ are needed.

PR #32, opened 2018-07-19
On notation for key combinations in agda2-mode

Kwezan:
**This is general remark to all chapters of Part I: Logical Fundations.**

The use of  **^C ^C** or **^C ^R** notation may suggest you are pressing the **Ctrl-Shift-c Ctrl-Shift-c** or **Ctrl-Shift-c Ctrl-Shift-r** key sequence, and this is not consistent with the generally accepted keystroke notation described in the Agda documentation (https://agda.readthedocs.io/en/v2.5.4/tools/emacs-mode.html)  and implemented in the emacs editor (agda2-mode).
In addition, the meaning of **Ctrl-b** and **Ctrl-B** key sequences is different  in the emacs editor. The former means "move backward one character" and the latter means "move backward and highlight one character". The same applies to the other key combinations **Ctrl-F**, **Ctrl-P** and **Ctrl-N**.

### DIFF
--- a/src/plfa/Naturals.lagda
+++ b/src/plfa/Naturals.lagda
@@ -738,7 +738,7 @@ Begin by typing
     m + n = ?
 
 The question mark indicates that you would like Agda to help with filling
-in that part of the code. If you type `^C ^L` (control-C followed by control-L)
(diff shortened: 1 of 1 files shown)


wadler:
Thank you! You are quite right, I should use the same notation as the Agda docs.

PR #3, opened 2018-01-31
fixed typos.

jonaprieto:
Thanks for writing this. I'm just reading and I found some mistakes. I hope it helps. :)

PR #37, opened 2018-07-20
fix Agda command notation #3

Kwezan:


wadler:
Thank you! I appreciate you paying attention to small details, such as the case of letters in commands.

Kwezan:
Thank you for your warm words.
Because I really like this book, I would like to help as much as I can to make it free as possible of all sorts of mistakes.

PR #38, opened 2018-07-20
fix type signatures

Kwezan:


wadler:
Yes, but the submitted code doesn't have the answers (nor should it, because this is an exercise). I have updated the exercises to contain postulates (thanks to you), so am closing this now. Thank you!

PR #42, opened 2018-07-25
Use name instead of handle in acknowledgements

potato4444:


wadler:
Thank you for the correction!

PR #43, opened 2018-07-25
add new Unicode name

Kwezan:


wadler:
Thanks for fixing my omission!

PR #59, opened 2018-09-06
use --local-references to build book

Cubesoup:
possibly the least interesting pull request.


wenkokke:
Yeah, I'll do this myself 😂 Specifically, this breaks because Travis CI doesn't have the latest version of `agda2html` yet.

PR #60, opened 2018-09-06
Use new local reference syntax

Cubesoup:
Book builds and links work (including #sections). 

I assume that the link named More on line 729 of Untyped.lagda is intended to point to plfa.More, not plfa.Lambda, and have fixed this.


wadler:
Thank you!

PR #62, opened 2018-09-08
Add 'circled plus' unicode character to Lists

spwhitt:


spwhitt:
Just realized this character is only used in the text of the lesson, not in the code. So I guess adding it to the list is not really necessary.

PR #65, opened 2018-09-21
Update Naturals.lagda

Kwezan:
Currently the second interference rule starts form fifth line of the definition (including empty line between rules).
I propose to simplify the relevant part of the text.


wenkokke:
I'm merging the proposed change from the text, but not directly from the pull-request due to the included changes of `---` to `--`.

PR #74, opened 2018-10-02
Font stuff

Cubesoup:
Added instructions for making nice unicode in Emacs on DICE machines.

Changed Fonts.lagda as per discussion w/ Phil.


wenkokke:
Any reason we're suggesting the use of DejaVu Sans Mono instead of FreeMono?

PR #89, opened 2018-11-08
Remove subst-split

laMudri:
In this development, subst-split ended up just being a complicated way of getting hold of the linearity of matrix multiplication (that Δ(Γ₁ + Γ₂) = ΔΓ₁ + ΔΓ₂). I think it's still needed in my quantitative development because the underlying linear algebra is built intrinsically into resourced substitution, rather than being separated out. The extrinsic version here is far clearer.

### DIFF
--- a/src/plfa/Linearity.lagda
+++ b/src/plfa/Linearity.lagda
@@ -626,56 +626,6 @@ weaken-⊢₁ {γ} {Γ} {A} {B} {M} ⊢₁M
 \end{code}
 
 \begin{code}
-subst-split : ∀ {γ δ} (Γ Γ₁ Γ₂ : Context γ)
-  → (σ : ∀ {A} → γ ∋ A → δ ⊢ A)
-  → (Δ : ∀ {A} → γ ∋ A → Context δ)
-  → (P : ∀ {A} → (x : γ ∋ A) → Δ x ⊢₁ σ x)
-  → Γ ≡ Γ₁ ++ Γ₂
-  → Σ[ Δ₁ ∈ (∀ {A} → γ ∋ A → Context δ) ]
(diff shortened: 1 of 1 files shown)


laMudri:
Just to put it somewhere, here is a big list of linear algebra readings for that file:

`_**_` – scaling
`**-identity` – unit scaling does nothing
`**-assoc` – scaling by a product is the composition of scalings

`0∙` – the 0 vector
`0∙-absorb` – scaling the 0 vector gives the 0 vector
`0**Γ≡0∙γ` – scaling by 0 gives the 0 vector

`_++_` – vector addition
`++-identityˡ/ʳ` – adding the 0 vector does nothing
`++-comm` – vector addition is commutative
`++-assoc` – vector addition is associative
`**-distribʳ-++` – scaling by a sum gives the sum of the scalings
`**-distribˡ-++` – scaling a sum gives the sum of the scalings

`VC i` – ith standard basis vector
`smash Γ Δ` – matrix-vector multiplication ΔᵀΓ
`smash-0∙` – linear maps preserve the 0 vector
`smash-0#` – adding a row of 0s to the end of the matrix and then multiplying by a vector produces a vector with a 0 at the bottom
`smash-**` – linear maps preserve scaling
`smash-++` – linear maps distribute over sums
`smash-VC₁` – multiplying by a standard basis vector projects out the corresponding column of the matrix
`smash-VC₂` – the standard basis vectors put together give the identity matrix

`rename-⊢₁` – we can do a change of basis (or, in this case, permutation of basis and introduction of 0-size dimensions) on the substitution matrix
`weaken-⊢₁` – we can introduce one 0-size dimension
`subst-⊢₁` – we can actually use the substitution matrix

PR #92, opened 2018-11-20
add Google Seattle course

halfaya:


halfaya:
Thank you! My web page for the course is internal to Google, but honestly it's not very exciting. We're following the textbook and I'm suggesting some exercises for the students to concentrate on trying. I also recommended they watch your excellent talk on PAT at Strange Loop. We meet once a week for an hour Monday 1-2, and there are about 6 students total, but who shows up and who does any work outside that hour varies depending on how busy everyone is.

I'm hoping to apply Agda to our work, for more precise specification of our APIs and functionality as well as obtaining better correctness guarantees, even if this just means automatic generation of tests from a spec. So the goal of the class is to get other people up to speed to understand what I'm doing and be able to help out. If all goes well perhaps I can write up an "experience report" for a future ICFP.

Thanks so much for the creation of this wonderful course! Very timely. A year or two before I'd actually gone through the first several chapters of SF and redid all the exercises in Agda--I had a very similar experience to you, being surprised that I preferred the Agda solutions to the Coq ones almost without exception. I really appreciate that you put in the effort to rework the course, and I find I'm learning a lot from it as well.

PR #95, opened 2018-11-28
Fix formatting: Remove unintended list item

andrejtokarcik:
Search for the string "n + m is justified by" in the HTML rendition of the [chapter on Equality](https://plfa.github.io/Equality/#rewriting-expanded).

You'll notice an unexpected bullet point starting in the middle of the sentence as well as incorrectly rendered preformatted blocks in the surrounding text.

It is apparently caused by a plus sign left over as the first character on a line in the source, making the Markdown formatter interpret it as a new list item. The following changes should fix it.


andrejtokarcik:
As far as I can grep through the sources, this is the only occurrence of such a leading plus sign.

PR #101, opened 2018-12-02
Don't conflate "term" with "value"

iblech:
In my understanding, there are many distinct *terms* of type ⊤, for instance `tt`, `id tt`, `id (id tt)` and others not of this family. But there is only one *value* (by which I mean the elements in some unspecified, maybe informal, denotational model).

In case that you feel that the terminology is different, just close this pull request. I know that this is slightly slippery terrain.


wenkokke:
@wadler What are your thoughts on this?

PR #104, opened 2018-12-02
Use https instead of http where possible

iblech:
Not trying to start a war, just had a couple seconds to spare. All remaining `http` links are either below `assets/` or their `https` analogues don't work. Merge or close as you see fit.


wenkokke:
Yup, agreed 👍

PR #110, opened 2018-12-04
Update Naturals.lagda

Kwezan:
I propose to apply the following convention in order to unify the text:
- if the code follows a full sentence and refers to it, then use a colon
- if the code is in the middle of a sentence, do not use a colon and possibly use other punctuation marks depending on the context

### DIFF
--- a/src/plfa/Naturals.lagda
+++ b/src/plfa/Naturals.lagda
@@ -21,7 +21,7 @@ left over as you started with.
 
 ## The naturals are an inductive datatype
 
-Everyone is familiar with the natural numbers:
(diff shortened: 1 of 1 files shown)


wadler:
Thank you. You made me rethink some of my choices. For the various "Stories of Creation", I edited to put a colon before the display, and made the leading comment into a full sentence. If you could make a pull request for similar changes in the next chapter, Induction, I would be grateful.

Kwezan:
Suer

Kwezan:
Yes, I will do.

PR #113, opened 2018-12-05
Update Induction.lagda

Kwezan:
the continuation of the agreed convention on unification of the text


Kwezan:
> the continuation of the agreed convention on unification of the text

Kwezan:
It was my mistake

wadler:
Thank you!

PR #116, opened 2018-12-06
Fixes the character sequence for ℕ in Emacs

mdimjasevic:
The character sequence for obtaining the symbol for the set of natural numbers in Emacs is incorrect. The chapter on natural numbers says it is `\bN` when in fact it is `\bn`. The Verified Functional Programming in Agda book also says it is `\bn`.

### DIFF
--- a/src/plfa/Naturals.lagda
+++ b/src/plfa/Naturals.lagda
@@ -921,7 +921,7 @@ Information on pragmas can be found in the Agda documentation.
 
 This chapter uses the following unicode:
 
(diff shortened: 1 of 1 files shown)


wenkokke:
It's `\bN`, see: 
https://github.com/agda/agda/blob/master/src/data/emacs-mode/agda-input.el#L488
It was changed a while back, to allow for more consistent names for blackboard bold letters, e.g. using `\bn` for 𝕟.

PR #123, opened 2018-12-07
Last Update Lists.lagda

Kwezan:
This is my last adjustment at the moment.
Before I proceed with the revision of Part 2, I have to read it carefully several times.


Kwezan:
Ok, I agree with you. I will continue the update of Part 2.

PR #128, opened 2018-12-10
Update Properties.lagda

Kwezan:
Please consider distinguishing the `gas` variable from the "gas" name in Evaluation section.


wadler:
I reviewed this section, and changed one use of `gas` to _gas_. The rest looks in order to me.

PR #135, opened 2018-12-13
Update Inference.lagda

Kwezan:
continuation of improving the consistency of the text

### DIFF
--- a/src/plfa/Inference.lagda
+++ b/src/plfa/Inference.lagda
@@ -123,7 +123,7 @@ another that takes it as an input.
 ## Synthesising and inheriting types
 
 In addition to the lookup judgment for variables, which will remain
-as before, we now have two judgments for the type of the term.
+as before, we now have two judgments for the type of the term:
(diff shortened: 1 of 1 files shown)


src/plfa/Inference.lagda
```diff
@@ -506,8 +506,8 @@ _≟Tp_ : (A B : Type) → Dec (A ≡ B)
 ...  | yes refl | yes refl  =  yes refl
 \end{code}
 
-We will also need a couple of obvious lemmas: the domain
-and range of equal function types are equal.
+We will also need a couple of obvious lemmas; the domain
```
wenkokke: I'm not so sure about this usage of a semicolon.

Kwezan: or you can change to "We will also need a couple of obvious lemmas that the domain
and range of equal function types are equal."

wenkokke:
Accepting this pull request as-is.

PR #140, opened 2018-12-17
Make usage of well-founded consistent

k4rtik:


k4rtik:
Oh, you are right, that makes sense! I think I was concerned because "well-founded" is used as a technical term in PL literature, so perhaps it should have a consistent form in prose.

I learned that there doesn't seem to be a clear consensus on the issue of hyphenation: https://english.stackexchange.com/a/430858

PR #141, opened 2018-12-18
Does wordsmithing in the chapter on relations

mdimjasevic:
This patch does wordsmithing in the chapter on relations:

1) it uses "the `where` clause" instead of "a `where` clause", and
2) it uses "on the left-hand side" instead of "of the left-hand side".


wenkokke:
Manually made the change to use "on the left-hand side" instead of "of the left-hand side", not copying the change to use "the where clause" instead of "a where clause"

PR #162, opened 2019-01-22
Isomorphism: aligns variable names in text with that in code

mdimjasevic:
In the chapter on isomorphism, this patch aligns variable names in text with that in code: I believe it should have been `y` instead of `x` if the text is to follow the code.


wadler:
Thanks for drawing my attention to this! Turns out, I used the phrase "left inverse" incorrectly. I've fixed the text now.

mdimjasevic:
@wadler , I hope you noticed there is "right inverse" in the nearby text as well.

PR #179, opened 2019-01-29
Connectives: clarifies what simplification a pattern match enables

mdimjasevic:
Just like in the pull request #177, in the chapter on connectives this patch replaces "equation" with "propositional equality" to clarify where simplification happens.

### DIFF
--- a/src/plfa/Connectives.lagda
+++ b/src/plfa/Connectives.lagda
@@ -258,8 +258,8 @@ value of type `⊤` must be equal to `tt`:
 η-⊤ tt = refl
 \end{code}
 The pattern matching on the left-hand side is essential.  Replacing
-`w` by `tt` allows both sides of the equation to simplify to the
-same term.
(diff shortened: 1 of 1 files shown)


wenkokke:
The use of the phrase "propositional equality" here seems cumbersome and confusing, I'd argue that this may not be a clarification.

mdimjasevic:
@wenkokke, this is exactly the same situation as in the PR #177, which Philip agreed with and approved.

The thing is that simplification happens only on the right-hand side of the equation, and on both sides of the propositional equality from the conclusion of the theorem, hence the proposed change. Therefore, this is actually more a correction than a clarification.

wenkokke:
You're right, let's go with consistency then, and I'll get over my feelings about moving from "equation" to "propositional equality".

wadler:
Nice attention to detail. May well help some student. Thank you!

PR #180, opened 2019-01-29
Connectives: fixes a typo, once -> one

mdimjasevic:
In the chapter on connectives, this patch changes "once member" to "one member" in a sentence:

"And, indeed, type `⊤` has exactly once member, `tt`."


wadler:
Thank you!

PR #184, opened 2019-01-30
Connectives: fixes an incorrect verb plural

mdimjasevic:
In the chapter on connectives, this patch fixes an incorrect verb plural.


wadler:
Thank you!

PR #185, opened 2019-01-30
Connectives: fixes an explanation of simplification

mdimjasevic:
Just like in the pull requests #177 and #179, in the chapter on connectives this patch replaces "equation" with "propositional equality" to better explain where simplification enables finishing a proof.


wadler:
Thank you!

PR #186, opened 2019-01-30
Connectives: changes a punctuation before a code block

mdimjasevic:
In the chapter on connectives, this patch replaces an occurrence of "as follows." with "as follows:". This is the only place in the whole `src` directory where a full stop was used after "as follows", unlike a colon that was used for 28 times.


wadler:
Thank you!

PR #189, opened 2019-01-31
Getting Started: pins down software versions so badges match links

mdimjasevic:
On the [Getting Started](https://plfa.github.io/GettingStarted/) page, there is a slight mismatch between badges on the top and corresponding links in the text:

- The Agda badge indicates version 2.5.4.2, while the text bellow links to the latest version. At the moment the latest Agda version is 2.5.4.2, but this can get out of sync.
- The standard library badge indicates version 0.17, while the text bellow links to the front page of the library repository, which might have changes that are not in version 0.17.

This patch pins down, i.e., fixes links to versions of Agda and the standard library so that they match versions indicated by the badges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ or by downloading [the zip archive](https://github.com/plfa/plfa.github.io/archi
     git clone https://github.com/plfa/plfa.github.io 
 
 Finally, we need to let Agda know where to find the standard library.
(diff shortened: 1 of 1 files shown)


wenkokke:
@wadler Yes, this is what we want. We want to have the book compile with specific versions of Agda, and we want to encourage *ourselves* to always keep the book working with the most recent versions.

PR #196, opened 2019-02-15
Naturals: clarifies a sentence about built-ins

mdimjasevic:
In the first chapter on naturals, this patch introduces a clarification in a sentence on built-ins by changing the usage of prepositions.


wenkokke:
While I think that you're right in pointing out the original sentence is confusing, I don't think your clarification actualy, uh, _clarifies_ it. I'll ponder for a bit on how to rewrite this sentence.

wadler:
Thank you!

PR #198, opened 2019-02-17
Minor alignment change.

ywata:
Though it's not a real issue, appearance will get consistent with this change.


ywata:
Yes. I should have added one more space.

PR #199, opened 2019-02-19
Quantifiers: fixes spacing in expressions

mdimjasevic:
In the chapter on quantifiers this patch adds and removes spacing in expressions where this was needed for consistency with the rest of the text.


mdimjasevic:
I thought that I should submit these two changes separately, but then I lumped them together.

I force-pushed here so that only extra space inside the braces is removed. That should meet your comment.

mdimjasevic:
And now I just force-pushed again to reintroduce the missing spacing in the Σ-syntax.

PR #208, opened 2019-02-28
Negation: adds a missing preposition

mdimjasevic:
In the chapter on negation, this patch adds a missing preposition "in".


mdimjasevic:
I do my best, but I am not a native English speaker and therefore not all my edits end up being the best or even meaningful proposals. That's why I rely on the pull request mechanism and expect situations like this one where you reject my edit. Hopefully not only my knowledge of computer science foundations improves as I work through the book, but my knowledge of English too.

wenkokke:
@mdimjasevic Your help is greatly appreciated!

PR #211, opened 2019-03-03
Properties, DeBrujin, More: Missing unicode sections added

ywata:
The three chapters does not have unicode section, I added unicode sections for them.
As I did not collect unicode characters manually, I implemented some perl script to collect in the following criteria.

- If any unicode character that is appeared in the current chapter is picked up in unicode section, if the character is appeared at most once in any preceding chapters.

I sorted the character based on unicode code point.

### DIFF
--- a/src/plfa/DeBruijn.lagda
+++ b/src/plfa/DeBruijn.lagda
@@ -1351,3 +1351,16 @@ The relation between the two approaches approximates the
 golden ratio: raw terms with a separate typing relation
 require about 1.6 times as much code as inherently-typed terms
 with de Bruijn indices.
+
+## Unicode
+
+This chapter uses the following unicode:
+
+    σ  U+03C3  GREEK SMALL LETTER SIGMA (\Gs or \sigma)
+    ₀  U+2080  SUBSCRIPT ZERO (\_0)
+    ₃  U+20B3  SUBSCRIPT THREE (\_3)
+    ₄  U+2084  SUBSCRIPT FOUR (\_4)
+    ₅  U+2085  SUBSCRIPT FIVE (\_5)
+    ₆  U+2086  SUBSCRIPT SIX (\_6)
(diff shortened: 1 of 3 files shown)


wenkokke:
Thanks so much. Would you consider committing your perl script as well?

ywata:
@wenkokke 
I'd be happy to provide the scripts. I'd like to know where I put files to do so.

What I created are:
- A text file that contains the list of files to be examined. I.e, Naturals.lagda, Induction.lagda ... in chapter order.
- A text file that contains all the instructions for unicode text, that are extracted from .lagda unicode section and added missing ones.
- A perl script that scans each .lagda file and generate all the unicode instructions for the file with .txt.
- A perl script that scans the above .txt file in chapter order to remove instructions that appeared more than twice and generate a file with .tx2.
- A bash script to run the perl scripts.

What I'd like to consider are:
- location of text files:  plfa.github.io/src/plfa
- location of perl and bash scripts: plfa.github.io/bin
- location of bash script to run in: plfa.github.io/src/plfa
- how to report if some unicode instruction is missing, currently report this to stderr and continue to run
Please let me know your opinion. I'll modify the scripts accordingly.

Another thoughts:
If you mark the position of unicode instructions in .lagda files, generating the unicode instruction can be automated. I tried to do this, but I'm not succeeded in this because I failed to run make build on my Mac, though I install gnu-sed.

wenkokke:
Oof, that's a bit more involved regarding intermediate files than I'd like. Could you file an issue to let me know what's going wrong with `make build`? I'm developing on macOS, so it's a bit troubling that you're experiencing errors on that platform.

ywata:
@wenkokke 
As the files are not so big and processing can be run on memory,
 I'll combine my perl script into one to avoid intermediate file.

For the first step, I'll generate unicode instruction files in text format.
Give me several days for my spare time. After I implement it, I'll raise a ticket to discuss about location etc.

Issue #210, opened 2019-03-02
More: Code block for exercise in More.lagda


ywata:
To apply proposed change in #143, I put code blocks with simple script before reading the contents (#181).
Now I'm reading More.lagda and I'm wondering adding code block at the end of More.lagda might not be what is originally intended or not.
I'd like to know how we solve the exercise in More.lagda.
There are at least two possibilities:
(a).  Add code, ex. code for sums,  in existing definition. 
(b).  Copy all the existing code to the end of the section after 'Your code goes here'. And start adding code for exercises.

(a) is easy and (b) introduces naming issue.
If (a) is what originally intended, I'd like to propose to remove the code block I previously added.
If (b) and using inner module is intended, I'd like to propose to include
`  module exercise where` line in the code block.

wenkokke:
@ywata Yeah, that code block should be removed, as the intention is to edit the entirety of More.lagda.

wadler:
Yes, that's right. Thanks, guys!

PR #212, opened 2019-03-03
More' Removed code block that is not need in More.lagda.

ywata:
Based on #210, I removed the code block in More.lagda.

Though I removed the code block, some instruction to add code for exercise in the existing code might helpful for readers. Because I was not certain how I solved the exercise in the beginning.

Instead of removing the code block, it might be better to change the comment to suggest how to solve. Please feel free to update this pr.


wenkokke:
Somehow, this branch has build conflicts.

wadler:
Thank you both!

PR #214, opened 2019-03-04
Properties: Unnecessary character removed.

ywata:
This is my mistake in my pr #211, which I found during unicode instruction generator implementation.


wenkokke:
Thanks.

PR #218, opened 2019-03-09
Decidable: uses a more specific relation symbol

mdimjasevic:
In the chapter on decidables, this patch makes usage of a more specific relation symbol to make it clear that it is the `_≤ᵇ_` function that is talked about instead of the relation `_≤_`.


mdimjasevic:
Any thoughts on this one?

wenkokke:
Actually, yes, this is fine. Sorry it took so long. Thanks @mdimjasevic!

PR #224, opened 2019-03-10
Decidable: removes unused and unrelated imports

mdimjasevic:
In the chapter on decidables, this patch removes unused and unrelated imports. For example, the chapter does not talk about lists at all so there is no need to import list constructors and list functions. The chapter type-checks just fine after removing these imports.

### DIFF
--- a/src/plfa/Decidable.lagda
+++ b/src/plfa/Decidable.lagda
@@ -23,18 +23,16 @@ of a new notion of _decidable_.
 
 \begin{code}
 import Relation.Binary.PropositionalEquality as Eq
-open Eq using (_≡_; refl; sym; trans; cong)
+open Eq using (_≡_; refl)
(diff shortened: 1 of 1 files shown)


wenkokke:
Is this import still needed?

```
- open import Relation.Nullary.Negation using (contraposition)
+ open import Relation.Nullary.Negation using ()
```

mdimjasevic:
Wen, I'd say it is because the import statement spans to the next line:

```agda
open import Relation.Nullary.Negation using ()
  renaming (contradiction to ¬¬-intro)
```

The `¬¬-intro` function is used in the chapter. I don't know if there's a better way to import only one function and rename it right away. I couldn't find a way to drop the `using ()` part.

wenkokke:
Nope, you're right.

PR #233, opened 2019-03-16
Reorder statistics page to have the same order as Table of Contents

googleson78:


wenkokke:
The statistics file was massively out of date, as atm it isn't generated automatically. I've updated the config file, which was the real cause of the inconsistency, and pushed those changes, together with the updated statistics, instead.

Issue #213, opened 2019-03-04
make build failed


ywata:
During my trial, I encountered two issues. I resolved one  but the other not.

- I had to include `standard-library` in `~/.agda/defaults`
- Currently, I have the following error

```
agda2html --verbose --link-to-local-agda-names --use-jekyll=out/ --link-to-agda-stdlib -i src/tmp/Quantitative.lagda -o out/tmp/Quantitative.md 2>&1 \
                | sed '/^Generating.*/d; /^Warning\: HTML.*/d; /^reached from the.*/d; /^\s*$/d'
/Users/test/work/plfa.github.io/src/tmp/Quantitative.lagda:57,8-25
The name of the top level module does not match the file name. The
module plfa.Quantitative should be defined in one of the following
files:
  /private/var/folders/k2/1jkk2gts1xn0scns9ym3r6m00000gn/T/agda2html10320/plfa/Quantitative.agda
  /private/var/folders/k2/1jkk2gts1xn0scns9ym3r6m00000gn/T/agda2html10320/plfa/Quantitative.lagda
  /Users/test/lang/agda/agda-stdlib/src/plfa/Quantitative.agda
  /Users/test/lang/agda/agda-stdlib/src/plfa/Quantitative.lagda
  /Users/test/work/plfa.github.io/src/tmp/plfa/Quantitative.agda
  /Users/test/work/plfa.github.io/src/tmp/plfa/Quantitative.lagda
  /Users/test/lang/agda/agda-2.5.4.2/.stack-work/install/x86_64-osx/lts-12.16/8.4.4/share/x86_64-osx-ghc-8.4.4/Agda-2.5.4.2/lib/prim/plfa/Quantitative.agda
  /Users/test/lang/agda/agda-2.5.4.2/.stack-work/install/x86_64-osx/lts-12.16/8.4.4/share/x86_64-osx-ghc-8.4.4/Agda-2.5.4.2/lib/prim/plfa/Quantitative.lagda
agda --allow-unsolved-metas --html '--html-dir=/private/var/folders/k2/1jkk2gts1xn0scns9ym3r6m00000gn/T/agda2html10320' '--include-path=/Users/test/work/plfa.github.io/src/tmp' /Users/test/work/plfa.github.io/src/tmp/Quantitative.lagda
```

wenkokke:
Ah! I'm pretty sure I know what's going on – `agda2html` isn't used to dealing with parameterised modules!

ywata:
[...]

PR #235, opened 2019-03-21
Mention jekyll as a build dependency

ywata:
During #213 investigation, I noticed installing jekyll is needed to build the book. It's better to mention it in README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ To build and host a local copy of the book, there are several tools you need *in
   - [agda2html](https://github.com/wenkokke/agda2html)
   - [Ruby](https://www.ruby-lang.org/en/documentation/installation/)
   - [Bundler](https://bundler.io/#getting-started)
+  - [Jekyll](https://jekyllrb.com/)
   
 For most of the tools, you can simply follow their respective build instructions.
 Most recent versions of Ruby should work.


wenkokke:
Jekyll is used to build and host the book, but it is automatically installed using bundler.

ywata:
Ah. I see. In my case,I had to install it explicitly. I have one possible cause of this and I'll check it.

ywata:
@wenkokke 

I checked the possible issue. It was not cause of my case. 
After I uninstall everything bundle related, I tried to run make again. 
It gave me'
```make
ruby -S bundle exec jekyll build
bundler: command not found: jekyll
Install missing gem executables with `bundle install`
make: *** [build] Error 127
```

I read Makefile and I guess you expect us to run make macos-setup before running make build on MacOS.

wenkokke:
That's a good point, there should be a general setup command. Thanks.

ywata:
As https://jekyllrb.com/ says,
installing bundler and jekyll is possible with the command below:
`gem install bundler jekyll`
This was easy for me.

wenkokke:
The correct way to install all the ruby dependencies is to call:
```bash
gem install bundler
bundle install
```

PR #237, opened 2019-03-22
more symbols on the font test page. removed somw symbols not used in book.

Cubesoup:


Cubesoup:
The extra \rho / \Rho thing was a mistake -- good catch! 

The extra lambda symbol was present further down the list, but I've moved it next to the normal lambda (it's occurrence as four symbols, not as part of the alphabet). 

I've moved the symbols that are egregiously non-monospaced (on my screen, on Github) to the section at the bottom.

wenkokke:
@Cubesoup did you check whether they ligned up (or not) on the font we use on the webpage, or using the font you use on your editor/system? This looks good to me, btw.

Cubesoup:
I checked with my editor (which uses a bunch of different fonts at the moment), and with the Github font. I can't currently build the book locally, so it would be difficult for me to check with the webpage font.

In any case, I think the font test page is more aspirational than anything right now.

wenkokke:
Point taken.

Cubesoup:
Done.

Cubesoup:
I have added "fake" long arrows to the font test page (em-dash followed by an arrow symbol).

PR #247, opened 2019-04-06
Lambda: removes a mention of a non-existing name

mdimjasevic:
In the chapter on lambda calculus, this patch removes a mention of the non-existing name `W` in the definition of the `Value` type constructor.


wadler:
I updated Lambda to clarify.

mdimjasevic:
You're welcome!

PR #272, opened 2019-05-31
add another missing import

tammela:


wenkokke:
Could you clarify in what sense these are missing?

tammela:
> Could you clarify in what sense these are missing?
Missing `≲` definition for exercise "`—↠≲—↠′`" (https://github.com/plfa/plfa.github.io/pull/271)
Missing `×` definition for exercise "``Context-≃`"

wenkokke:
I don't think we should add imports for constructs that are needed only during exercises. You, @wadler?

wadler:
@tammela: Good catch, thank you. @wenkokke: Good question. I lean toward leaving it to the student to add such imports when they are required. @tammela, let us know if you think that puts too big a burden on the students.

tammela:
@wenkokke @wadler seems reasonable to let the students import whatever they need.
Thanks for the clarification

PR #275, opened 2019-06-03
Lambda: fixes typing symbols in a context

mdimjasevic:
In the chapter on lambda calculus, this patch fixes the symbol used for types in a typing context (a plain colon replaced with `\:` in the Agda input method).


wenkokke:
Thanks, @mdimjasevic!

PR #277, opened 2019-06-04
Lambda: adds a connective for better readability

mdimjasevic:
In the chapter on lambda calculus, this patch adds the "and" connective after a bullet point for improved readability of the whole sentence.


wenkokke:
Honestly, I don't have any strong feelings about this; I think the version without "and" is slightly cleaner.

PR #278, opened 2019-06-04
Lambda: fixes the description of the lookup judgment

mdimjasevic:
In the chapter on lambda calculus, this patch fixes the description of lookup judgment. A variable with the same name can be shadowed only by a latter element in the list, and not by by an element that came before it.

### DIFF
--- a/src/plfa/Lambda.lagda
+++ b/src/plfa/Lambda.lagda
@@ -1058,7 +1058,7 @@ The constructors `Z` and `S` correspond roughly to the constructors
 `here` and `there` for the element-of relation `_∈_` on lists.
 Constructor `S` takes an additional parameter, which ensures that
 when we look up a variable that it is not _shadowed_ by another
-variable with the same name earlier in the list.
+variable with the same name latter in the list.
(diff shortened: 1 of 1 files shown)


wenkokke:
The patch you submitted says "latter" instead of "later". Furthermore, contexts are built up from right-to-left, so earlier might be more accurate than later?

mdimjasevic:
> The patch you submitted says "latter" instead of "later".

My English is not perfect so I might be wrong here, but I thought that "latter" is more appropriate here. Should it be "later" instead?

> Furthermore, contexts are built up from right-to-left, so earlier might be more accurate than later?

Oh, I believe you're right here. I confused lists for contexts, so all is good. I'm closing this.

odanoburu:
damn English can be confusing https://english.stackexchange.com/questions/164073/what-is-the-difference-between-later-and-latter

wadler:
Thank you for the attention to detail. We'll leave it as is for now, but I think there is a case to replace 'earlier' by 'to its left'.

PR #279, opened 2019-06-04
Lambda: removes a superfluous backtick in a typing judgment

mdimjasevic:
In the chapter on lambda calculus, this patch removes a superfluous backing in a typing judgment.


wenkokke:
Thanks!

PR #282, opened 2019-06-05
Lambda: fixes the description of lookup judgment

mdimjasevic:
Related to the closed pull request #278, this replaces 'earlier' by 'to its left' so let me know if that would work now.


mdimjasevic:
> Thank you. I said "to its left" before, but I now believe I should have said "to its right". Please confirm, and if so update accordingly.

No, it is "to its left" in the list. It would be "to its right" in the context. The non-empty constructor for list appends to the opposite side compared to the non-empty constructor for context. Therefore, this pull request needs no changes.

mdimjasevic:
@wadler, please let me know if my justification sounds good to you and if this patch can be merged without changes.

mdimjasevic:
Dear Philip, this is a gentle reminder of this pending patch. As explained above, I believe there are no further changes to be done and the patch is ready for merging. Can you please confirm?

PR #283, opened 2019-06-05
Properties: Fixed italics and capitalization

h4iku:
The italic style wasn't closed after the book name. Also, the word "chapter" was capitalized in previous chapters when referring to a specific one. I'm not sure about capitalizing "volume".


h4iku:
Updated.

PR #293, opened 2019-06-15
DeBruijn: Replaced Ax with  ⊢` in the second example

h4iku:
Replaced `Ax` with  `` ⊢` `` in `⊢plus`, and fixed the markdown style of some variable names.


wadler:
Thank you!

PR #295, opened 2019-06-15
Replaced occurrences of 'De Bruijn' with 'de Bruijn'

h4iku:
There was a mixture of using 'De Bruijn' and 'de Bruijn' in Chapter DeBruijn and some other parts of the book, I replaced all of them with 'de Bruijn' except the ones at the beginning of the sentences.


wadler:
Thank you.

PR #298, opened 2019-06-19
[ upgrade ] Agda 2.6.0.1 and (upcoming) stdlib 1.1

gallais:
We were talking about SPLV yesterday and realised it would be nice to have Agda
installed on the machines in the lab for both Phil & Conor's lectures. The sensible
solution seems to synchronise on the latest released version.

The changes needed to upgrade are pretty light. The fact we have managed to get
rid of `Data.String.Unsafe` raises the question of how much of PLFA can be checked
with the `--safe` option turned on.

Note that this PR is not ready yet:

* [ ] Agda stdlib 1.1 has not been released yet (agda/agda-stdlib#808)
* [ ] I have not looked into updating the build system to pick up the new versions

You should be able to push additional commits to the branch to update the build
system before merging.

### DIFF
--- a/src/plfa/Inference.lagda
+++ b/src/plfa/Inference.lagda
@@ -251,8 +251,7 @@ import Relation.Binary.PropositionalEquality as Eq
 open Eq using (_≡_; refl; sym; trans; cong; cong₂; _≢_)
 open import Data.Empty using (⊥; ⊥-elim)
 open import Data.Nat using (ℕ; zero; suc; _+_)
-open import Data.String using (String)
-open import Data.String.Unsafe using (_≟_)
+open import Data.String using (String; _≟_)
 open import Data.Product using (_×_; ∃; ∃-syntax) renaming (_,_ to ⟨_,_⟩)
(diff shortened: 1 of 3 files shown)


wenkokke:
If you look at the 2.6.0 branch of the repository, you'll note that I've done much of the work already. 

There's many more problems to overcome. The biggest problem is that `agda2html` processes the raw output of `agda --html`, and is broken quite severely by the upgrade. In response to this, we are moving away from `agda2html`, and are hoping to use the new in-place highlighter.

wenkokke:
As a matter of planning, Phil is currently teaching from the book, and as long as we don't have a stable URL (which we are also planning) we don't want to upgrade the main version of the book.

The current plan is to perform several upgrades in mid-July, if we don't encounter any serious problems. We plan to update to Agda 2.6.0.1 with the matching standard library, and to introduce a stable version of the book, hosted on [plfa.github.io/stable/](plfa.github.io/stable/), so that we can more aggressively update the main version of the book.

PR #303, opened 2019-06-20
Lambda: fixes a prose comparison in the primed exercise

mdimjasevic:
In the introductory chapter on lambda calculus, this patch fixes a comparison in prose of `` ` "x" `` and `"x"` (instead of just `x`). Therefore, the only difference is in the back-tick.

### DIFF
--- a/src/plfa/Lambda.lagda
+++ b/src/plfa/Lambda.lagda
@@ -212,7 +212,7 @@ definition may use `plusᶜ` as defined earlier (or may not
 
 #### Exercise `primed` (stretch)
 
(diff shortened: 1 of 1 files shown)


mdimjasevic:
> No, the proper comparison is with `x`, not `"x"`, because the update lets you write the variable x, which is itself bound to `` ` "x" ``.

Clearly I don't get that example. I couldn't unify the abbreviated form with the example and from the example I thought it could only be `"x"`; then I was asking myself what use there is in jumping through all the hoops just to write two characters less: `"x"` instead of `` ` "x"``.

I'll try reading it a few more times in hope it will eventually click.

PR #309, opened 2019-06-22
Lambda: small-step semantics and stepping to

mdimjasevic:
In the introductory chapter on lambda calculus, this patch adds a paragraph linking the introduction of the chapter where small-step semantics is mentioned ("This chapter formalises the simply-typed lambda calculus, giving its syntax, small-step semantics [...]") with the reduction relation. Without it there is no explanation that this is what is small-step semantics.

Furthermore, per discussion we had via email, this also explains what it means to step, which is needed for quiz questions following the paragraph.


mdimjasevic:
It reads really well now. Thank you for adding that paragraph!

PR #311, opened 2019-06-24
Lambda: removes unused imports

mdimjasevic:
In the introductory chapter on lambda calculus, this patch removes unused imports at the beginning of the chapter.


mdimjasevic:
> I don't import constructors unless I also import the relevant type. Can you update?

I just updated the patch by adding back the `Dec` type constructor. It should be ready for merging now.

PR #314, opened 2019-06-24
Lambda: makes a statement about what simply-typed lambda calculus is

mdimjasevic:
In the introductory on lambda calculus, this patch adds a statement explaining what formalises the simply-typed lambda calculus. In the introduction STLC is said to be formalised in the chapter, but nowhere else is this term used again. This patch fixes the issue.

### DIFF
--- a/src/plfa/Lambda.lagda
+++ b/src/plfa/Lambda.lagda
@@ -1151,6 +1151,9 @@ the three places where a bound variable is introduced.
 
 The rules are deterministic, in that at most one rule applies to every term.
 
+The syntax, operational semantics and typing rules as given by `Term`,
+`_—→_` and `_⊢_⦂_`, respectively, formalise the simply-typed lambda
+calculus.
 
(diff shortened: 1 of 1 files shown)


mdimjasevic:
Thank you for your feedback.

My point is that some concepts are mentioned in the chapter introduction, e.g., simply-typed lambda calculus: "This chapter formalises the simply-typed lambda calculus [...]", but without an explicit reference later in the text. By the end of the chapter it leaves me wondering which part of the chapter formalised STLC.

I can update the patch. Would it then be correct to say the following?

> The syntax, operational semantics and typing rules as given by `Term`, `_—→_` and `_⊢_⦂_`, respectively, formalise Plotkin’s Programmable Computable Functions.

mdimjasevic:
Dear @wadler, has this issue been addressed? By quickly scanning the chapter I don't see anything changed about this. Would you agree a sentence like I proposed in https://github.com/plfa/plfa.github.io/pull/314#issuecomment-506197093 would be good to have?

PR #315, opened 2019-06-24
Lambda: changes section levels to go in sequence

mdimjasevic:
In the introductory chapter on lambda calculus, this patch changes levels of some sections such that the hierarchy of sections goes step by step. For example, I changed it such that the first subsection of section 2.1 would be 2.1.1, and not 2.1.1.1. In other words, with this patch there is no level skipping. Also, some section level changes are done so they correctly reflect what (sub)section they belong to.


mdimjasevic:
@wenkokke, what was the resolution of this issue?

PR #335, opened 2019-07-09
Properties: adds an article in a section on substitution

mdimjasevic:
In the chapter on properties, this simple patch adds an article in the prose on substitution.


mdimjasevic:
I am not sure why this is closed. Is it because the article should not be there or did you forget this change after upgrading the book to Agda 2.6?

PR #348, opened 2019-07-16
Fix #347

cyberglot:
It fixes issue #347.

Tested on:
 - Mac
    + Firefox/mobile view @ latest;
    + Chrome/mobile view @ latest.
 - iPhone 8
    + Firefox @ latest;
    + Safari @ iOS 12.3.1.

PR #350, opened 2019-07-16
Patch/fix mobile menu

cyberglot:
Just like #348, but now `make test-offline` finishes without errors.


wadler:
Thank you! Leaving for Wen to confirm.

PR #356, opened 2019-07-18
Properties: fixes a typing judgment by replacing an identifier with a term

mdimjasevic:
In the chapter on properties, this simple patch fixes a typing judgment in prose by replacing an identifier with a term.


mdimjasevic:
You're right! I forgot the context that sets the scene for the `Z` constructor of the ni relation. I'm glad this is fixed now.

PR #373, opened 2019-08-05
[ fix ] bug in the last line of Formal vs. Informal

gallais:
In the last line of https://plfa.github.io/Lambda/#formal-vs-informal
we get a spurious link to the standard library because the script
happily matches "Agda."

Note that I couldn't test this as `make` fails with weird warnings
on my laptop.

### DIFF
--- a/highlight.sh
+++ b/highlight.sh
@@ -81,7 +81,7 @@ fi
 
 # Create a sed script which matches and replaces all Agda standard library links
 if [ ! -f "$AGDA_STDLIB_SED" ]; then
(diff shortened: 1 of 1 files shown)


wenkokke:
So that's what that was; I was wondering!

wenkokke:
Could you report the errors in make as an issue?

wenkokke:
Tested locally. Fixes this particular issue, but in the future we should probably move to matching only hrefs.

gallais:
> Could you report the errors in make as an issue?

I pushed in a rush yesterday afternoon. Now that I have actually installed the
dependencies, it works just fine.

PR #374, opened 2019-08-07
Properties: adds a sentence on canonical forms

mdimjasevic:
In the chapter on properties, this patch adds a sentence about canonical forms. I'm not completely sure if it makes a good statement, but with this I at least wanted to start a discussion on what are canonical forms; the prose no where says that at the moment.


mdimjasevic:
Thank you for this! I agree my proposal was not the best. Now I understand this better.

PR #375, opened 2019-08-08
the successor -> VM

kenichi-asai:
I think ''the successor'' in the original text refers to `(V-suc VM)` but it should be `VM`.


kenichi-asai:
Thank you for the reply.  I missed the presence of "of" in the original text.

PR #396, opened 2019-09-02
part1/Connectives: Add `iff` import

trajafri:
Exercise `⇔≃×` needs iff


trajafri:
Based on #272, this is okay

PR #401, opened 2019-09-11
Add unicode reference sidebar

qaisjp:
Hi.

I'm hoping to do this course this semester, and I'm trying to get ahead a little. I'm still on the first chapter, but I find myself either:
- opening another tab to read the "Unicode" section, or
- scrolling down to the bottom of the page to read the "Unicode" section

I find this a little bit inconvenient, so I've built a little sidebar that will extract and display the first codeblock following the "Unicode" header, within the next two next sibling elements, on any webpage that has the `page.layout == "page"`.

#### Developer/other notes
- This "sidebar" is sticky (using inbuilt `position: sticky;`, so no additional JavaScript libraries).
- The visibility is persisted to `localStorage`. This means that if you choose to show the reference, and then refresh the page (or open the website in another tab), the reference will still show.
- The default visibility is _hidden_.
- The decision to not use jQuery was deliberate.
- This was developed against the _Naturals_ chapter, and I can confirm it works against every page defined in _Part 1_ and _Part 2_. Other pages that may have a "Unicode" section were not tested.
- I tested "Dedication", "Preface", and "Acknowledgements" to see how it behaves against pages without the "Unicode" section — no "toolbar" is shown.

#### If this feature is of your interest, I will do further work to:

- ensure this works on evergreen browsers
- make it responsive (it currently behaves poorly at different sizes)
- ensure this works excellently (or is disabled) on mobile
    - I may leave mobile support for a future PR if this proves troublesome
- optimise and further document wherever possible
- remove debug lines

Below are some screenshots and a GIF. Your feedback is greatly appreciated!

#### Everything, in video format

NimbleSelfishGander

#### The page in its natural form

image

#### After clicking show reference

image

#### Hidden, whilst scrolled down

image

#### Shown, whilst scrolled down

image


wenkokke:
How would this scale on e.g. mobile? Did you test it across various browsers?

qaisjp:
> It's hard to tell from the images how much of the screen gets used for the Unicode reference.

Okay, so **there's still work to be done on small sizes** (see the rest of this comment).

I'm hoping these screenshots will help visualise.

**Super wide screen**

image

**Widescreen (1920x1080)**

image

**Small squarey monitor (1024x768)**

image

**Super low res**

image


> How would this scale on e.g. mobile?

Over the weekend I came across this page: https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme

It seems to solve a similar same issue that this PR attempts to solve.

I think drafts.csswg.org solves it elegantly & in a much better way than I have.

The csswg.org website behaves identically on mobile and desktop. Observe that the sidebar and "Jump to TOC" links show the same content, but in different ways. Observe that clicking the back button after "Jumping to TOC" will return you to the original location — I think this is default browser location.

I am thinking of simply replicating exactly that behaviour, but in this pull request. What do you think?

> Did you test it across various browsers?

Not yet. See _"If this feature is of your interest, I will do further work to"_ in my main comment.

I pledge to ensure support for the following platform/browser combinations:

- Firefox
   - 60 ESR, which IS maintains for Windows
   - whichever version DICE has
- Edge (whichever version IS maintains for Windows)
- Chrome
   - latest version on macOS
   - whichever version IS maintains for Windows
   - whichever version DICE has
- Safari (latest version on macOS)

This covers all of the browsers that the [university officially supports](https://www.ed.ac.uk/information-services/computing/application-development/application-support/supported-web-browsers), with the following caveats:
- `IE 11` has been excluded
- `macOS / Safari` has been added
- Windows 7 specifically will not be tested, only Windows 10 (this probably will not matter)

I don't really mind making sure it works on IE 11... but plz no. Up to you.

> On Desktop, it's fairly easy to keep a second tab open at the Unicode section.

Easy, yes. Although it's still a bit inconvenient. There's the issue of having to decide between switching tabs and switching windows (different keybinds). That issue can be solved on some platforms by using three windows. On macOS it isn't solved because `cmd+tab` switches between apps, and ``cmd+` `` switches between windows of the current app. So there's still two different keyboard shortcuts there.

Yeah, maybe I'm a little lazy. But still "lazy" enough to submit a pull request that attempts to optimise that pain away.

> Plus, this problem may be better solved in Emacs lisp rather than on the website.

Agreed.

> I don't want to end up having to maintain any (more) JavaScript for this project.

I get it. JavaScript is a crappy language, and I hate it as much as the next person. I know your point isn't about "language hate", but I'm just putting it out there.

Let me know if there's anything I can do to improve documentation or make things clearer.

Additionally, if you still just simply don't want to have to deal with this code, I won't be disheartened if you simply don't want this code in the repo. I'll just maintain a fork or release a greasemonkey script or something, and put it on [Better Informatics](https://betterinformatics.com), so that others can benefit from the work.

wenkokke:
I think the best solution would be for you to publish this as a GreaseMonkey scipt. That way, no additional JavaScript code ends up in the repository. We'd be happy to link to the script from the "Getting Started" page.

PR #402, opened 2019-09-16
Add missing library file

momirza:
Adding this missing file fixes the import issue:
```
The name of the top level module does not match the file name. The
module Assignment1 should be defined in one of the following files:
...
```


wenkokke:
Thanks, @momirza!

PR #406, opened 2019-09-17
[WIP] Restructure Lambda now that String equality computes

wenkokke:
Starting agda-stdlib version 1.0 the equality on strings computes during type checking. This makes it possible to simplify a lot of the [Lambda](https://plfa.github.io/Lambda/) chapter. Specifically, we can get rid of the unsightly [`_≠_`](https://plfa.github.io/Lambda/#35615).

I would removed the ticked syntax exercise, as this never really added much. This frees Lambda from any use of the `impossible` technique, and would move the explanation to `lookup` in [DeBruijn](https://plfa.github.io/DeBruijn/#12879), the next serious usage.

- [X] remove `_≠_`
- [X] replace instances of `_≠_` with `(λ())`
- [x] move text cautioning the use of postulating `impossible` the next instance of this pattern

Tests pass, I think this is ready to merge.

### DIFF
--- a/src/plfa/part2/DeBruijn.lagda.md
+++ b/src/plfa/part2/DeBruijn.lagda.md
@@ -423,7 +423,7 @@ lookup ∅       _        =  ⊥-elim impossible
 We intend to apply the function only when the natural is
 shorter than the length of the context, which we indicate by
 postulating an `impossible` term, just as we did
-[here]({{ site.baseurl }}/Lambda/#impossible).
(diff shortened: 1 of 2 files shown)


wenkokke:
I'm also working on a nicer way to write successors in Lambda, but I figured I'd keep the changes minimal.

wadler (approved):
Looks good! Thank you.

PR #411, opened 2019-09-19
Revising Bin

wadler:
Changed from prefix bits to postfix bits, which puts the least-significant bit on the right (as one would expect) rather than on the left.


qaisjp:
Can we pick which way to use for the assignment, or _must_ we use the old way for assignment 1?

PR #423, opened 2019-10-11
inductive data declaration for isomorphism typo

citrusmunch:
assuming `record _≃_` should be independent from `data _≃′_`, the latter should have primes on all the functions


wenkokke:
The names `to`, `from`, etc for `_≃′_` are bound only in the datatype declaration, shadowing the top-level projections exported from the record. Therefore, this isn't an issue.

PR #424, opened 2019-10-15
Don't hardcode executable paths

mgttlinger:
`bash` is not specified to reside under `/bin/bash` on all OSs. E.g. on NixOS it does not.


wenkokke:
Thanks, @mgttlinger!

PR #438, opened 2019-11-13
Avoid line breaks in the middle of code

qaisjp:
**Before**

image

**After**

image

### DIFF
--- a/_sass/agda.scss
+++ b/_sass/agda.scss
@@ -38,6 +38,7 @@ pre.highlight {
 }
 code {
     @include code-font;
+    white-space: nowrap;
 }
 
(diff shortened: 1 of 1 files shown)


qaisjp:
Breaks multiline code blocks

wadler:
Thank you. Can you explain how the change breaks things? It makes sense that CSS would have a directive to discourage breaking quoted code in the middle (as in your example) and it would be great to take advantage of it.

qaisjp:
Agreed. I planned to reopen this when I had a suitable fix, which I've now pushed.

**Before (without fix)**
image

**After (with fix)**
image

PR #440, opened 2019-11-16
Prettify renamings by defining subset relation for contexts

pedrominicz:
Considering that the book tries to make every formula simple to read, I believe that the definition of this small auxiliary function is well worthwhile.

I would also like to define another function for substitutions, but I cannot think of a suitable name for it. In my notes, I called it `_⊑_` to give a hint about how it resembles `_⊆_`. An example of its use can be seen here: https://gist.github.com/pedrominicz/27c27aa491f981b6c2eebd64f5137348

I believe I have seen functions similar, if not equal, to `_⊑_` defined somewhere in the literature, but I could not find it. Name suggestions are welcome.


pedrominicz:
I am sorry for the `Merge pull request #1 from plfa/dev` commit in this pull request. I am sure there is a way to update my fork/create pull requests. I will make sure to do better in the future.

wadler:
Thanks for the suggestion! It is nice that this makes the statement so much shorter.  But it will take some work to make clear the relation between a renaming map and a substitution map. I don't have time to think it through in the next few weeks, but might return to it for winter break. Leaving this issue open for now.

wenkokke:
@wadler Do you think you will return to this, or can I close this pull request?

PR #442, opened 2019-12-28
Lambda: give the suc constructor is the text with an argument

mdimjasevic:
In the chapter on lambda calculus, this simple patch changes how the `suc` constructor is given in the text. A consistent way in the introduction of the chapter would be: `suc` is given an argument, `zero` takes no arguments and `case` takes 4 arguments.


mdimjasevic:
Please feel free to squash commits when merging the pull request.

PR #453, opened 2020-01-24
Properties: visual alignment of pattern matches for M —→ N

mdimjasevic:
In the chapter on properties, this patch visually aligns pattern-matches for `M —→ N` in the proof of `preserve`. Of course, the proof type-checks without the parentheses, but having them is more visually appealing by bringing it in line with the rest of the pattern-matches, including with that of `(β-μ)` where parentheses are not mandatory either.


mdimjasevic:
You are very much welcome!

PR #454, opened 2020-01-24
Properties: introduce an implicit argument M for clarity in the type of M-named variables

mdimjasevic:
In the chapter on properties, this simple patch introduces an implicit argument `M` for clarity in the type of M-named variables in the proof of `eval`. The patch makes the proof more pedagogical and easier to write with Agda's interactive type-driven development.

Without the patch, the types of variables are confusing, e.g., the variable `M—↠N` has type `N —↠ N₁`. For the following code with a hole:

```agda
eval : ∀ {L A}
  → Gas
  → ∅ ⊢ L ⦂ A
    ---------
  → Steps L
eval {L} (gas zero)    ⊢L                              =  steps (L ∎) out-of-gas
eval {L} (gas (suc m)) ⊢L with progress ⊢L
... | done VL                                          =  steps (L ∎) (done VL)
... | step L—→M with eval (gas m) (preserve ⊢L L—→M)
...    | steps M—↠N fin                                =  {!!}
```

the goal and context are:

```
Goal: Steps L
————————————————————————————————————————————————————————————
fin  : Finished N₁
M—↠N : N —↠ N₁
N₁   : Term  (not in scope)
m    : ℕ
L—→M : L —→ N
N    : Term  (not in scope)
⊢L   : ∅ ⊢ L ⦂ A
A    : Type  (not in scope)
L    : Term
```

When the patch is applied, the context changes such that the variables with "M" in their names also have M in the corresponding place in their types:

```
Goal: Steps L
————————————————————————————————————————————————————————————
fin  : Finished N
M—↠N : M —↠ N
N    : Term  (not in scope)
m    : ℕ
L—→M : L —→ M
M    : Term
⊢L   : ∅ ⊢ L ⦂ A
A    : Type  (not in scope)
L    : Term
```

PR #459, opened 2020-01-28
Properties: fix the implementation of eval

mdimjasevic:
In accordance with issue #455, in the chapter on properties this patch fixes the theorem and the proof for `eval`. I made the changes as small as possible.

I used the `⊢L` name in the type constructor `Steps`, although the next chapter simply says "Steps L" instead of `Steps ⊢L` (though in my opinion the latter would be more appropriate).

The implicit variable `M` is there per https://github.com/plfa/plfa.github.io/pull/454#pullrequestreview-348936479. I kept it implicit because making it explicit would require a change in the definition of the `Progress` type constructor and in all usage sites too.

Your feedback would be highly appreciated.

### DIFF
--- a/src/plfa/part2/Properties.lagda.md
+++ b/src/plfa/part2/Properties.lagda.md
@@ -931,27 +931,27 @@ Given a term `L` of type `A`, the evaluator will, for some `N`, return
 a reduction sequence from `L` to `N` and an indication of whether
 reduction finished:
 ```
-data Steps (L : Term) : Set where
+data Steps {L A} : ∅ ⊢ L ⦂ A → Set where
 
-  steps : ∀ {N}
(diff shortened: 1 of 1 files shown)


mdimjasevic:
> What do you think? Would you be willing to enter a second pull request, that makes the changes above (and also in subsequent chapters) and rewrites the text in Properties to align with the new design?

Your version of `Steps` for Chapter Properties definitely makes more sense than my version! I realized only after reading your version that with my version of `Steps` in Chapter Properties we have two potentially different values of type `∅ ⊢ L ⦂ A`, which is not a good thing.

I will make a second pull request per your suggestion for Chapter Properties and subsequent chapters.

wenkokke:
I'm closing this, as you mentioned making a new pull request, @mdimjasevic.

PR #460, opened 2020-01-30
Add note about Agda pragma placement

zachrbrown:
Hello! Firstly, thank you for writing this book — I am excited to work through it!

Secondly, the reason for this pull request! Coming from Haskell, I am used to placing my LANGUAGE pragmas at the top of my file, and ended up upsetting Agda when trying to do the same when working through the Naturals chapter. It wasn't clear to me when working through the chapter that the pragma statements actually look for previously defined bindings within the module (and additionally that the built-ins can be bound to any name — neat!). It became clear once I read [the docs](https://agda.readthedocs.io/en/v2.5.4.1/language/built-ins.html#natural-numbers) that this is indeed the case. 

Hoping this minor addition to the section on using the built-in pragmas can help future readers.

### DIFF
--- a/src/plfa/part1/Naturals.lagda.md
+++ b/src/plfa/part1/Naturals.lagda.md
@@ -239,11 +239,10 @@ Including the line
 ```
 tells Agda that `ℕ` corresponds to the natural numbers, and hence one
 is permitted to type `0` as shorthand for `zero`, `1` as shorthand for
-`suc zero`, `2` as shorthand for `suc (suc zero)`, and so on.  The
-declaration is not permitted unless the type given has exactly two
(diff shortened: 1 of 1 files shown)


zachrbrown:
Makes sense — it is getting a bit into the weeds of Agda specifics; thus diverting from the focus of the book, as you mentioned. Your suggestion sounds good to me. I'll make the recommended adjustment. Thanks for taking a look!

wenkokke:
You've written `succ` instead of `suc` ^__^"

zachrbrown:
@wenkokke oh no, whoops! Thanks for catching that.

PR #464, opened 2020-02-13
Correct wrong book title

MatthiasGabriel:
Correct wrong book title. The book is called Begriffsschrift.


wenkokke:
Yup, it is! Thanks!

PR #465, opened 2020-02-13
Remove strange line break

MatthiasGabriel:
Remove a strange line break in the middle of the formula. I guess the mark-down respects that line break because it's part of the formula.
image


wenkokke:
Thanks, @MatthiasGabriel!

PR #470, opened 2020-04-21
proposed text addition: pragmatic note on matching against decidable equality

peterthiemann:
The first formalization of the lambda using strings for variables has at least one pretty subtle point, which I probably overlooked in the text. For example, it matters a lot whether you write  y ≟ x or  x ≟ y. In one case, the proofs go through nicely, but in the other case Agda dumps a screenful of messages from the low level implementation of string comparison on you.


peterthiemann:
I wouldn't start the paragraph with **note that**, but it's ok otherwise.

PR #474, opened 2020-05-20
Fix cong-rename

L-TChen:
`M` on the right hand side of `≡` should be `M′`.


jsiek:
I've gone ahead an made a further simplification to cong-rename after this fix, which is to
remove the premise M ≡ M′. I think that premise was a workaround for some
temporary problem I was dealing with that no longer exists.

PR #481, opened 2020-06-17
EPUB: Fix broken links

mreed20:
This brings us down to 10 `epubcheck` warnings. This is done with a lua filter that rewrites links with regular expressions.

References #480.

### DIFF
--- a/epub/rewrite-links.lua
+++ b/epub/rewrite-links.lua
@@ -1,0 +1,43 @@
+local currentChapter = nil
+
+-- Reassigns identifiers for all all Headers level 2 and higher. Level 2 Headers
+-- correspond to chapters, and are identified by the first word in their content
+-- field. Headers of level more than 2 are identified by "#<chapter>-<anchor>",
+-- where "<chapter>" is the identifier of the chapter this header is nested in,
+-- and "<anchor>" is this Header's existing identifier.
+function Header (el)
+    if el.level == 2 then
+        local title = pandoc.utils.stringify(el.content[1])
+        currentChapter = title:match("%w+")
+        el.identifier = currentChapter
+    elseif el.level > 2 then
+        el.identifier = currentChapter .. '-' .. el.identifier
+    end
+    return el
+end
+
+
+
+-- Performs the following transformations on Link targets:
+--
+--     Case 1:
+--         [text]({{ site.baseurl }}/chapter/#more-stuff) -> [text](#chapter-more-stuff)
+--
+--     Case 2:
+--         [text]({{ site.baseurl }}/chapter/)            -> [text](#stuff)
+--
+-- All other Links are ignored.
+function Link (el)
+    local n
+    -- When parsing a markdown link such as "[stuff]({{ site.baseurl }}/Naturals",
+    -- pandoc encodes the link's URL with percent-encoding, resulting in the
+    -- link "[stuff](%7B%7B+site.baseurl+%7D%7D%2FNaturals)". This is why the
+    -- pattern begins with ".*site%.baseurl.*" and not "{{ site%.baseurl }}".
+    el.target, n = el.target:gsub("^.*site%.baseurl.*/(%w+)/#([%w-]+)$",  -- case 1
+                                  "#%1-%2")
+    if n == 0 then
+        el.target = el.target:gsub("^.*site%.baseurl.*/(%w+)/$",          -- case 2
+                                "#%1")
+    end
+    return el
+end
--- a/epub/rewrite-links.lua
+++ b/epub/rewrite-links.lua
@@ -24,19 +24,19 @@ end
 --         [text]({{ site.baseurl }}/chapter/#more-stuff) -> [text](#chapter-more-stuff)
 --
 --     Case 2:
---         [text]({{ site.baseurl }}/chapter/)            -> [text](#stuff)
+--         [text]({{ site.baseurl }}/chapter/)            -> [text](#chapter)
 --
 -- All other Links are ignored.
 function Link (el)
     local n
     -- When parsing a markdown link such as "[stuff]({{ site.baseurl }}/Naturals",
     -- pandoc encodes the link's URL with percent-encoding, resulting in the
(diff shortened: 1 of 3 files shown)


src/plfa/part2/More.lagda.md
```diff
@@ -1234,7 +1234,7 @@ side to be well typed.
 
 ## Test examples
 
-We repeat the [test examples]({{ site.baseurl }}/DeBruijn/#examples) from Chapter [DeBruijn]({{ site.baseurl }}/DeBruijn),
+We repeat the [test examples]({{ site.baseurl }}/DeBruijn/#examples) from Chapter [DeBruijn]({{ site.baseurl }}/DeBruijn/),
```
mreed20, line 1237: This is done so the regex in the lua filter's regex matches. It's already done for all other links.

wenkokke:
Your comment says:

```lua
-- Performs the following transformations on Links:
--
--     Case 1:
--         [text]({{ site.baseurl }}/stuff/#more-stuff) -> [text](#more-stuff)
--     Case 2:
--         [text]({{ site.baseurl }}/stuff/)        -> [text](#stuff)
--
-- Links that do not match the forms above are kept as is.
```

Isn't that the wrong rewrite? It seems this would send `/DeBruijn/` to `#DeBruijn`, which is wrong, and `/DeBruijn/#11370` to `#11370` which is also wrong?

mreed20:
The tag `#DeBruijn` works because of the following code in `rewrite-links.lua`:

```lua
-- Add identifiers to chapter headings (level 2). The identifier is based on the
-- 'permalink' field of that chapter's YAML metadata block.
function Header (el)
    if el.level == 2 then
        local title = pandoc.utils.stringify(el.content[1])
        el.identifier = title:match("^(%w+):")
    end
    return el
end
```

which transforms headers of the form

```
# DeBruijn: other stuff
```

to the form

```
# DeBruijn: other stuff {#DeBruijn}
```

The header above can now be referenced from any markdown document included in the EPUB (not just `DeBruijn.lagda.md`) like so: `[this is a link to DeBruijn](#DeBruijn)`. This works because pandoc essentially concatenates all the markdown files together when making the EPUB.

Did I address your concern?

mreed20:
By the way, when mapping `/DeBruijn/#11370` to `#11370`, I'm assuming that there isn't another link with the name `/DeBruijn/#11370` (which would also map to `#11370`).

wenkokke:
> This works because pandoc essentially concatenates all the markdown files together when making the EPUB.

Ah! I suspected as much. Not incredibly neat, but I suppose it works.

> By the way, when mapping /DeBruijn/#11370 to #11370, I'm assuming that there isn't another link with the name /DeBruijn/#11370 (which would also map to #11370). This simplifies the link transformation since we don't need to know which chapter #11370 is in.

I don't know for sure, but I don't think we can guarantee this? However, I'd have to investigate.

wenkokke:
Oh, right. You said:
> This works because Pandoc essentially concatenates all the markdown files together when making the EPUB.
However, that's done in `include-files.lua`, isn't it? That script is already generating the chapter headings:
```
-- Before shifting headings, add a title heading at the beginning of the chapter.
local heading = pandoc.Header(1, pandoc.Str(pandoc.utils.stringify(document.meta.title)))
document.blocks:insert(1, heading)
```
At this point, we have both the filename and the chapter name. Therefore, we could traverse all anchors in the chapter and prefix them with the chapter name, e.g., `name="<anchor>"` to `name="<chapter>-<anchor>".

We then wouldn't have to worry about collisions in the IDs assigned by Agda, and just traverse all the links in the document and map `{{ site.baseurl }}/<chaper>/#<anchor>` to `#<chapter>-#<anchor>`.

mreed20:
> However, that's not something Pandoc does out of the box, right?

I'm not totally sure, but I believe pandoc does do it out the box. This article [Making an ebook with pandoc](https://pandoc.org/epub.html).

> At this point, we have both the filename and the chapter name. Therefore, we could traverse all anchors in the chapter and prefix them with the chapter name, e.g., name="<anchor>" to `name="-".
>
> We then wouldn't have to worry about collisions in the IDs assigned by Agda, and just traverse all the links in the document and map {{ site.baseurl }}/<chaper>/#<anchor> to #<chapter>-#<anchor>.

That's what I wanted to avoid doing 😛 . But now that you describe its benefits, it does sound like a more robust solution. I'll look into that!

mreed20:
@wenkokke Just pushed my changes. I'm now generating header identifiers the way you recommended.

epub/rewrite-links.lua
```diff
@@ -0,0 +1,43 @@
...
+
+
+
+-- Performs the following transformations on Link targets:
+--
+--     Case 1:
+--         [text]({{ site.baseurl }}/chapter/#more-stuff) -> [text](#chapter-more-stuff)
+--
+--     Case 2:
+--         [text]({{ site.baseurl }}/chapter/)            -> [text](#stuff)
```
wenkokke, line 27: This should probably be `s/stuff/chapter/`?

epub/rewrite-links.lua
```diff
@@ -0,0 +1,43 @@
...
+--
+-- All other Links are ignored.
+function Link (el)
+    local n
+    -- When parsing a markdown link such as "[stuff]({{ site.baseurl }}/Naturals",
+    -- pandoc encodes the link's URL with percent-encoding, resulting in the
+    -- link "[stuff](%7B%7B+site.baseurl+%7D%7D%2FNaturals)". This is why the
+    -- pattern begins with ".*site%.baseurl.*" and not "{{ site%.baseurl }}".
```
wenkokke, line 35: It would probably be a wee bit neater to explicitly match the %7Bs and %2Fs?

mreed20: good idea, I suppose my laziness has been found out 😄

mreed20:
Just pushed the new changes, thanks for the super quick review!

wenkokke:
Thanks, @mreed20!

PR #484, opened 2020-06-20
EPUB output: Populate acknowledgements.md with contributors

mreed20:
See the commit message for more information.

References #480.

## TODO

- [x] bracketed inclusion of `<ul class="list-of-contributors"> ... </ul>`
- [x] deal with `site.contributors`
- [x] include `liquid.lua`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ Gemfile.lock
 
 ## Emacs files
 auto/
+
+## Misc build files
+out/
+src/plfa/acknowledgements_epub.md
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ addons:
     packages:
       - libgmp-dev
       - libicu-dev
+      - luarocks
 
 # Ensure we run BASH and not SH
 env:
--- a/Makefile
+++ b/Makefile
@@ -49,12 +49,10 @@ out/:
 #       sections are displayed to users. Some readers may be slow if the chapter
 #       files are too large, so for large documents with few level-1 headings, one
 #       might want to use a chapter level of 2 or 3."
-#
-#TODO: embedded fonts not working (path problem?)
 
 epub: out/plfa.epub
 
-out/plfa.epub: out/ $(AGDA) $(LUA) epub/main.css
+out/plfa.epub: out/ $(AGDA) $(LUA) epub/main.css src/plfa/acknowledgements_epub.md
 	pandoc --strip-comments \
 		--css=epub/main.css \
 		--epub-embed-font='assets/fonts/mononoki.woff' \
@@ -62,6 +60,7 @@ out/plfa.epub: out/ $(AGDA) $(LUA) epub/main.css
 		--epub-embed-font='assets/fonts/DejaVuSansMono.woff' \
 		--lua-filter epub/include-files.lua \
 		--lua-filter epub/rewrite-links.lua \
+		--lua-filter epub/rewrite-html-ul.lua \
 		--lua-filter epub/default-code-class.lua -M default-code-class=agda \
 		--standalone \
 		--fail-if-warnings \
@@ -70,7 +69,8 @@ out/plfa.epub: out/ $(AGDA) $(LUA) epub/main.css
 		-o "$@" \
 		epub/index.md
 
-
+src/plfa/acknowledgements_epub.md: src/plfa/acknowledgements.md _config.yml 
+	lua epub/run-liquid.lua _config.yml $< > $@
 
 
 # Convert literal Agda to Markdown
@@ -119,7 +119,7 @@ build-incremental: $(MARKDOWN)
 
 # Remove all auxiliary files
 clean:
-	rm -f .agda-stdlib.sed .links-*.sed
+	rm -f .agda-stdlib.sed .links-*.sed src/plfa/acknowledgements_epub.md
 ifneq ($(strip $(AGDAI)),)
 	rm $(AGDAI)
 endif
@@ -158,6 +158,9 @@ travis-setup:\
 	$(HOME)/agda-stdlib-$(AGDA_STDLIB_VERSION)/src\
 	$(HOME)/.agda/defaults\
 	$(HOME)/.agda/libraries\
+	$(HOME)/.local/share/lua/5.1/tinyyaml.lua\
+	$(HOME)/.local/share/lua/5.1/liquid.lua\
+	$(HOME)/.local/share/lua/5.1/cjson\
 	/usr/bin/pandoc
 
 .phony: travis-setup
@@ -203,6 +206,17 @@ $(HOME)/.local/bin/agda:
 	unzip -qq $(HOME)/agda-$(AGDA_VERSION).zip -d $(HOME)
 	cd $(HOME)/agda-$(AGDA_VERSION);\
 		stack install --stack-yaml=stack-8.0.2.yaml
+
+$(HOME)/.local/share/lua/5.1/tinyyaml.lua:
+	luarocks install lua-tinyyaml
+
+$(HOME)/.local/share/lua/5.1/liquid.lua:
+	luarocks install liquid
+
+$(HOME)/.local/share/lua/5.1/cjson:
+	# Only this particular version works:
+	# https://github.com/mpx/lua-cjson/issues/56:
+	luarocks install lua-cjson 2.1.0-1
 
 travis-uninstall-agda:
 	rm -rf $(HOME)/agda-$(AGDA_VERSION)/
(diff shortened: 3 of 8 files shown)


wenkokke:
Changing the layout by removing the list-of-contributors class isn't an option. What is the HTML generated by Pandoc? Is there a way to get the class back in?

wenkokke:
On second thought: your current approach adds yet another dependency to building the web version of the book (i.e. Pandoc) and it may be a good idea to avoid that. Adding Lustache as a dependency for building the EPUB version may be an acceptable compromise.

mreed20:
> Changing the layout by removing the list-of-contributors class isn't an option. What is the HTML generated by Pandoc? Is there a way to get the class back in?

Got it. I didn't expect to keep that class removed; it was just to temporarily quell the epubcheck warnings. I'll look into ways around that. We can probably bracket the opening and closing `<ul>` tags with a template variable like `isWebsite`, which when defined to a non-null value would include the tags and exclude them otherwise.

> On second thought: your current approach adds yet another dependency to building the web version of the book (i.e. Pandoc) and it may be a good idea to avoid that. Adding Lustache as a dependency for building the EPUB version may be an acceptable compromise.

Wouldn't that still incur an additional dependency for the website generation (i.e., lustache)? Lustache's template syntax is different from the Jekyll template syntax used previously, so users would still need to install lustache to build the website.

mreed20:
Also, in the deleted `acknowledgements.md`, what's the difference between `site.contributors` and `site.extra_contributors`? I didn't see `site.contributors` defined in `_config.yml`.

wenkokke:
> Also, in the deleted `acknowledgements.md`, what's the difference between `site.contributors` and `site.extra_contributors`? I didn't see `site.contributors` defined in `_config.yml`, but I'm wondering if I should include it in the template file.

`site.contributors` is computed from GitHub data and appended to `_config.yaml` for deployed builds only, whereas `site.extra_contributors` is managed manually, and is for folks whose contributions were outside of GitHub or aren't registered by GitHub – due to, e.g., git misconfiguration.

wenkokke:
> > Changing the layout by removing the list-of-contributors class isn't an option. What is the HTML generated by Pandoc? Is there a way to get the class back in?
> 
> Got it. I didn't expect to keep that class removed; it was just to temporarily quell the epubcheck warnings. I'll look into ways around that. We can probably bracket the opening and closing `<ul>` tags with a template variable like `isWebsite`, which when defined to a non-null value would include the tags and exclude them otherwise.
> 
> > On second thought: your current approach adds yet another dependency to building the web version of the book (i.e. Pandoc) and it may be a good idea to avoid that. Adding Lustache as a dependency for building the EPUB version may be an acceptable compromise.
> 
> Wouldn't that still incur an additional dependency for the website generation (i.e., lustache)? Lustache's template syntax is different from the Jekyll template syntax used previously, so users would still need to install lustache to build the website.

Ugh, right... that's... annoying. I wish moustache templating was standardised. The easiest way to proceed is probably with a Ruby script which invokes Liquid, either as a shell script or as a Pandoc filter using [paru](https://heerdebeer.org/Software/markdown/paru/).

wenkokke:
It may be a good idea to translate your Lua code to Ruby... eventually... to reduce the number of scripting language we're depending on. So I'm not opposed to adding the dependency on paru.

mreed20:
> It may be a good idea to translate your Lua code to Ruby... eventually... to reduce the number of scripting language we're depending on. So I'm not opposed to adding the dependency on paru.

Yes, the number of languages is getting out of hand lol.

wenkokke:
Would you be able to translate your Lua scripts? Or should I plan on doing that myself?

mreed20:
> Ugh, right... that's... annoying. I wish moustache templating was standardised. The easiest way to proceed is probably with a Ruby script which invokes Liquid, either as a shell script or as a Pandoc filter using [paru](https://heerdebeer.org/Software/markdown/paru/).

Standard

mreed20:
Woah woops

wenkokke:
> Would you be able to translate your Lua scripts? Or should I plan on doing that myself?

For instance, there's an example here which includes files:
https://heerdebeer.org/Software/markdown/paru/#writing-and-using-pandoc-filters-with-paru#insert-other-pandoc-files

mreed20:
Sorry, wrong button. 

> Would you be able to translate your Lua scripts? Or should I plan on doing that myself?

I don't know ruby, so I'd prefer to avoid that for now. I don't mind learning it though. By the way, are you talking about the `format-acknowledgements` Lua script in particular, or also about the pandoc Lua filters? I ask because while pandoc supports writing filters in any language, their Lua filters are super fast: scroll down to the [end of this section](https://pandoc.org/lua-filters.html#introduction) for a performance comparison.

wenkokke:
Ohh, that's pretty neat, actually! I'm not too worried about the performance penality, since we're unlikely to be running this code very often. I'm more worried about the fact that I don't know Lua, and I'll likely have to maintain and extend this code. The fact that a Lua interpreter ships with Pandoc is reassuring, though.

mreed20:
Good point about the performance penalty---it doesn't matter much. And I completely understand the need to keep code maintainable.

> However, this does make me think... we've been wanting to move over to Hakyll & Pandoc for static site generation for a while now, so maybe we could eventually cut out Ruby altogether.

I'd love to look into this, although I don't know if I should do that before this is merged or after.

wenkokke:
For now, I wouldn't be too grumpy about you just straight-up including [liquid.lua](https://raw.githubusercontent.com/chenxianyu2015/liquid-lua/master/lib/liquid.lua) in the `epub/` directory, and loading it from there.

wenkokke:
> I'd love to look into this, although I don't know if I should do that before this is merged or after.

Yeah, probably after. Mostly, it means I'm not too upset about using Pandoc and some Lua scripts.

mreed20:
@wenkokke Just force pushed the new changes. Check out the commit message for more info. I also just did a fresh build of the website and ebook, and both are looking good.

wenkokke:
I'm getting the following error:
```
Error running filter epub/include-files.lua:
Error during function call: could not read argument 1: Expected an AST element, but could not parse value as such.
stack traceback:
	[C]: in function 'pandoc.utils.stringify'
	epub/include-files.lua:51: in function 'CodeBlock'
make: *** [out/plfa.epub] Error 83
```

mreed20:
Shoot, I couldn't reproduce that on my system when running `make epub` under a fresh install (via luarocks) of all the Lua dependencies and a clean git repository (via `git clean -fdx`).

What version of pandoc are you using? And what's the output of running `luarocks list`?

wenkokke:
Ah, see. I don't have luarocks, which probably means that this is a weird way for lua to tell me that I'm missing dependencies? Did you add instructions on installing the dependencies anywhere?

Mind you, this failure occurs while building the EPUB via `make`, not via `make epub`... I'm not sure if it should be included in the regular `make`?

mreed20:
> Ah, see. I don't have luarocks, which probably means that this is a weird way for lua to tell me that I'm missing dependencies? Did you add instructions on installing the dependencies anywhere?

Unfortunately, I think you're right. And good point: I'll add instructions to the README.

> Mind you, this failure occurs while building the EPUB via `make`, not via `make epub`... I'm not sure if it should be included in the regular `make`?

Sure, I can remove `epub` from the `all` Makefile target. I have no strong preference either way since most users will likely download the EPUB from Github, and not build it themselves.

mreed20:
Just pushed changes addressing your feedback. Lemme know how the instructions are for you.

wenkokke:
While I now have luarocks and the required packages, the error remains unchanged. I'm running Pandoc version 2.9.2.1.

mreed20:
What OS (+version) and Lua version are you using? I'd like to spin up a VM and try to reproduce this

wenkokke:
> What OS (+version) and Lua version are you using? I'd like to spin up a VM and try to reproduce this

MacOS 10.14 and Lua 5.3.5.

mreed20:
> MacOS 10.14 and Lua 5.3.5.

Same, but I'm on macOS 10.15.5. Could this be related to my force pushes?

wenkokke:
Maybe? Could you try building a fresh clone of the pull request?

mreed20:
aha! I reproduced it. Good call on the fresh clone. I also nuked my `~/.local/` directory and reinstalled Lua and Luarocks, which might have been related.

Will look into it now.

mreed20:
Check your `src/plfa/acknowledgements_epub.md`, which should be either
a) not have a definition for `title` in the metadata block (at the top of the file), or
b) be an empty file.

Either one of those possibilities triggered the same crash for me at `epub/include-files.lua:51`.

Alternatively, some other markdown file that is part of your build lacks the `title` attribute.

mreed20:
Just beefed up the error checking on `run-liquid.lua`, and I'm not experiencing the error on my end. Try doing a `make epub` from a fresh clone.

wenkokke:
It works now! I am still getting the error on nested p's, but I'm assuming that's due to the dedication.

wenkokke:
Thanks for all your work on this!

mreed20:
I should be thanking you, you've been committing nonstop today! I feel a little bad about all the Lua trouble I caused you with Travis; I'm glad the script is now in Ruby.

Issue #473, opened 2020-05-15
Why not Takahashi‘s method to prove confluence? 


L-TChen:
I was also trying to prove confluence using Takahashi’s approach, and with the sigma-algebra developed in the section on substitution I am able to prove the following `par-complete-dev` by a simple induction on the derivation of parallel reduction. 
```
_⁺ : Γ ⊢ A → Γ ⊢ A
(` x) ⁺       =  ` x
(λ̇ M) ⁺       = λ̇ (M ⁺)
((λ̇ M) · N) ⁺ = M ⁺ [ N ⁺ ]
(M · N) ⁺     = M ⁺ · (N ⁺)

par-complete-dev : M =⇒ N → N =⇒ M ⁺
```
Then the confluence property follows easily.

But, in the development version of plfa, I noticed the following paragraph 

> We opted not to use the “complete developments” approach of Takahashi (1995) because we felt that the proof was simple enough based solely on parallel reduction.

If the proof can be made even simpler, then why not?

jsiek:
Dear Liang-Ting Chen,

That's great that you've given this approach a try. 
My concern with Takahashi's method for this section is that I believe
there is a trade-off: one defines a second notion of parallel reduction
(an increase in complexity) in exchange for a proof that includes one
less case (decrease in complexity). So it was not obvious to me
that Takahashi's method would lead to a simpler development 
all things considered.

But now that you have done this alternative proof, we can compare the
two developments to see which is really simpler. Would you be
willing to share your Agda code?

Best regards,
Jeremy

L-TChen:
Sorry for my late reply. I don't bring my work laptop home (even though I still cannot resist thinking about it and working from a tablet), but my formalisation needs to be polished in the style of plfa. 

Here is the core part of my formalisation:
```agda
_⁺ : ∀ {Γ A}
  → Γ ⊢ A → Γ ⊢ A
(` x) ⁺       =  ` x
(ƛ M) ⁺       = ƛ (M ⁺)
((ƛ M) · N) ⁺ = M ⁺ [ N ⁺ ]
(M · N) ⁺     = M ⁺ · (N ⁺)

par-dev : ∀ {Γ A} {M N : Γ ⊢ A}
  → M ⇛ N
    -------
  → N ⇛ M ⁺
par-dev pvar = pvar
par-dev (pbeta M⇛M′ N⇛N′) =
  sub-par (par-dev M⇛M′) (par-dev N⇛N′) 
[...]

PR #493, opened 2020-07-14
[ re #473 ] Takahashi's complete development

L-TChen:
Regarding #473, this is my first try in reformulating the diamond property (triangle property, actually) using Takahashi's approach. I am not a very good storyteller, so the informal development might need to be polished further. Please feel free to improve my wording.

Thanks to Andrea @iwilare for the naming and the triangle diagram, I implemented them altogether. 

For the record, here are the details of my changes. 

* There are more high-level descriptions but less Agda code obviously. The main difference in terms of proofs between these two properties is that a single induction on `M ⇛ N` suffices to prove the triangle property instead of a double induction to prove the diamond property. The latter involves actually a number of cases eliminated implicitly by unification in Agda.

* The diamond property for the parallel reduction in this setting becomes optional. The triangle property is now used to prove the strip lemma without extracting the confluent term and the required parallel reduction explicitly, i.e. this step
```
    with par-diamond mn mm'
... | ⟨ L , ⟨ nl , m'l ⟩ ⟩
``` 
I let the original proof be an exercise, so the reader might appreciate Takahashi's approach.

* Oh, the last one,  the codepoint for the triple arrow ⇛ should be U+21DB instead of U+3015.

### DIFF
--- a/src/plfa/part2/Confluence.lagda.md
+++ b/src/plfa/part2/Confluence.lagda.md
@@ -610,9 +610,7 @@ and `par-confluence` are based on the notion of complete development
 by Takahashi (1995) and Pfenning's 1992 technical report about the
 Church-Rosser theorem. In addition, we consulted Nipkow and
 Berghofer's mechanization in Isabelle, which is based on an earlier
-article by Nipkow (JAR 1996).  There are many more mechanizations of
-the Church-Rosser theorem that we have not yet had the time to read,
-including Shankar's (J. ACM 1988) and Homeier's (TPHOLs 2001).
+article by Nipkow (JAR 1996).  
 
 ## Unicode
 
--- a/src/plfa/part2/Confluence.lagda.md
+++ b/src/plfa/part2/Confluence.lagda.md
@@ -447,9 +447,10 @@ par-triangle pvar          = pvar
 par-triangle (pabs p)      = pabs (par-triangle p)
 par-triangle (pbeta p1 p2) =
   sub-par (par-triangle p1) (par-triangle p2)
-par-triangle {M = (ƛ _) · _} (papp (pabs p1) p2) = pbeta (par-triangle p1) (par-triangle p2)
-par-triangle {M = ` _ · _}   (papp p1 p2) = papp (par-triangle p1) (par-triangle p2)
-par-triangle {M = _ · _ · _} (papp p1 p2) = papp (par-triangle p1) (par-triangle p2)
+par-triangle (papp {L = ƛ _ } (pabs p1) p2) =
+  pbeta (par-triangle p1) (par-triangle p2)
+par-triangle (papp {L = ` _}   p1 p2) = papp (par-triangle p1) (par-triangle p2)
+par-triangle (papp {L = _ · _} p1 p2) = papp (par-triangle p1) (par-triangle p2)
 ```
 
 The proof goes as follows.
--- a/src/plfa/part2/Confluence.lagda.md
+++ b/src/plfa/part2/Confluence.lagda.md
@@ -410,11 +410,24 @@ some `L`.  The typical proof is an induction on `M ⇛ N` and `M ⇛ N′`
 so that every possible pair gives rise to a witeness `L` given by
 performing enough beta reductions in parallel.
 
-Will it be even nicer if we simply perform as many beta reductions in
-parallel as possible on `M`, say `M ⁺`, so that `N ⇛ M ⁺` holds
-independent of another `N′`? This is exactly the idea of Takahashi's
-complete development.
- 
+However, a simpler approach is to perform as many beta reductions in
+parallel as possible on `M`, say `M ⁺`, and then show that `N` also
+parallel reduces to `M ⁺`. This is the idea of Takahashi's _complete
+development_. The desired property may be illustrated as
+
+        M
+       /|
+      / |
+     /  |
+    N   2
+     \  |
+      \ |
+       \|
+        M⁺
+
+where downward lines are instances of `⇛`, so we call it the _triangle
+property_.
+
 ```
 _⁺ : ∀ {Γ A}
   → Γ ⊢ A → Γ ⊢ A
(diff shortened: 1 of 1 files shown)


src/plfa/part2/Confluence.lagda.md
```diff
@@ -401,105 +401,108 @@ sub-par : ∀{Γ A B} {N N′ : Γ , A ⊢ B} {M M′ : Γ ⊢ A}
...
-* Suppose that `(ƛ N) · M ⇛ N₁ [ M₁ ]` and `(ƛ N) · M ⇛ N₂ [ M₂ ]`.
-  By the induction hypothesis we have
-  `N₁ ⇛ N₃` and `N₂ ⇛ N₃` for some `N₃`.
-  Likewise, we have
-  `M₁ ⇛ M₃` and `M₂ ⇛ M₃` for some `M₃`.
-  We choose `L = N₃ [ M₃ ]`.
-  We have both `(ƛ N₁) · M₁ ⇛ N₃ [ M₃ ]`
-  and `(ƛ N₂) · M₂ ⇛ N₃ [ M₃ ]`
-  by rule `pbeta`
+This step is optional, though, in the presence of triangle proeprty.
```
wenkokke, line 491: *proeprty -> property

src/plfa/part2/Confluence.lagda.md
```diff
@@ -605,19 +605,18 @@ Broadly speaking, this proof of confluence, based on parallel
...
+Church-Rosser theorem. In addition, we consulted Nipkow and
 Berghofer's mechanization in Isabelle, which is based on an earlier
-article by Nipkow (JAR 1996).  We opted not to use the "complete
-developments" approach of Takahashi (1995) because we felt that the
-proof was simple enough based solely on parallel reduction.  There are
-many more mechanizations of the Church-Rosser theorem that we have not
-yet had the time to read, including Shankar's (J. ACM 1988) and
-Homeier's (TPHOLs 2001).
+article by Nipkow (JAR 1996).  There are many more mechanizations of
+the Church-Rosser theorem that we have not yet had the time to read,
```
wenkokke, line 614: Not sure if this admission is warranted in a textbook?

jsiek:
I think we can remove that statement now.

-Jeremy

src/plfa/part2/Confluence.lagda.md
L-TChen: Thanks! Fixed.

src/plfa/part2/Confluence.lagda.md
L-TChen: Removed.

src/plfa/part2/Confluence.lagda.md
```diff
@@ -401,105 +401,108 @@ sub-par : ∀{Γ A B} {N N′ : Γ , A ⊢ B} {M M′ : Γ ⊢ A}
...
-some `L`.  The proof is relatively easy; it is parallel reduction's
-_raison d'etre_.
+some `L`.  The typical proof is an induction on `M ⇛ N` and `M ⇛ N′`
+so that every possible pair gives rise to a witeness `L` given by
+performing enough beta reductions in parallel.
+
+Will it be even nicer if we simply perform as many beta reductions in
+parallel as possible on `M`, say `M ⁺`, so that `N ⇛ M ⁺` holds
+independent of another `N′`? This is exactly the idea of Takahashi's
+complete development.
```
jsiek, line 416: The sentence structure is a bit surprising. How about the following instead:

However, a simpler approach  is to perform as many beta reductions in parallel as
possible on `M`, say `M ⁺`, and then show that `N` also parallel reduces to `M ⁺`.
This is the idea of Takahashi's *complete development*.

src/plfa/part2/Confluence.lagda.md
```diff
@@ -401,105 +401,108 @@ sub-par : ∀{Γ A B} {N N′ : Γ , A ⊢ B} {M M′ : Γ ⊢ A}
...
+    ︲\
+    ︲ \
+    ︲  \
+    ︲   N
+    ︲  /
+    ︲ /
+    ︲/
+    M⁺
+
+so that we call it the _triangle property_ which is proved by induction on `M ⇛ N` only.
```
jsiek, line 439: I would move this diagram up, just before the definition of `_⁺`, so that it comes right after
the intuitive description of complete development.

src/plfa/part2/Confluence.lagda.md
```diff
@@ -401,105 +401,108 @@ sub-par : ∀{Γ A B} {N N′ : Γ , A ⊢ B} {M M′ : Γ ⊢ A}
...
+  It follows `(λ L′) · M′ ⇛ L ⁺ [ M ⁺ ]`.
+
+* Suppose `x · M ⇛ x · M′`. By the induction hypothesis we have `M′ ⇛ M ⁺`
+  and `x ⇛ x ⁺` so that `x · M′ ⇛ x · M ⁺`.
+  The remaining case is proved in the same way, so we ignore it.  (As
+  there is currently no way in Agda to expand the catch-all pattern in
+  the definition of `_⁺` for us before checking the right-hand side,
+  we have to write down the remaining case explicitly.)
+
+The diamond property then follows by halving the diamond into two triangles.
```
jsiek, line 480: How about another diagram here to show the diamond with the two triangles?

src/plfa/part2/Confluence.lagda.md
L-TChen: Done. Thanks for rephrasing it!

src/plfa/part2/Confluence.lagda.md
L-TChen: Done.

src/plfa/part2/Confluence.lagda.md
L-TChen: Good idea. Done.

wenkokke:
@jsiek are you happy to merge this?

jsiek:
Yes, thanks!

PR #497, opened 2020-07-23
remove sentence calling proofs of negations proofs by contradiction

fredrikNordvallForsberg:
A minor thing, but the negation chapter calls _proofs of negations_ (assume `A`, derive `⊥`, hence `¬ A`) instead _proofs by contradiction_ (assume `¬ A`, derive `⊥`, hence `A`).

See eg [Andrej Bauer](http://math.andrej.com/2010/03/29/proof-of-negation-and-proof-by-contradiction/) arguing that one shouldn't conflate these two principles.

I couldn't think of a better fix than to simply remove the sentence, but I think the flow of the text still makes sense.

### DIFF
--- a/src/plfa/part1/Negation.lagda.md
+++ b/src/plfa/part1/Negation.lagda.md
@@ -34,8 +34,6 @@ as implication of false:
 ¬_ : Set → Set
 ¬ A = A → ⊥
 ```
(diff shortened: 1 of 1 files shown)


fredrikNordvallForsberg:
It's not cut-and-dried indeed, but I agree _reductio ad absurdum_ is better, thank you.

PR #512, opened 2020-09-07
[DeBruijn] Fixes a verb plural error

mdimjasevic:
In the DeBruijn chapter, this simple patch fixes a verb plural error.


wadler (approved):
Thank you!

Issue #513, opened 2020-09-09
Alternative formalisation of variables via de Bruijn indices


mdimjasevic:
In chapter "DeBruijn: Intrinsically-typed de Bruijn representation", there is a formalisation of variables `#_` via de Bruijn indices:

```agda
lookup : Context → ℕ → Type
lookup (Γ , A) zero     =  A
lookup (Γ , _) (suc n)  =  lookup Γ n
lookup ∅       _        =  ⊥-elim impossible
  where postulate impossible : ⊥

count : ∀ {Γ} → (n : ℕ) → Γ ∋ lookup Γ n
count {Γ , _} zero     =  Z
count {Γ , _} (suc n)  =  S (count n)
count {∅}     _        =  ⊥-elim impossible
  where postulate impossible : ⊥

#_ : ∀ {Γ} → (n : ℕ) → Γ ⊢ lookup Γ n
# n  =  ` count n
```

This formalisation relies on postulates to avoid handling the case when a de Bruijn index is out of context bounds. In my view, usage of postulates should be discouraged in the book as postulates in general open room for inconsistency.

Instead, I propose an alternative definition of the three functions above, `lookup'`, `count'` and `#'_`, respectively, that uses no postulates, yet the alternative functions guarantee that a de Bruijn index is within the bounds. This development should require no further modifications to the formalisation in the rest of the chapter. At first I was planning to propose to make `Context` indexed by its size, but I gave up on that approach as it would require quite a few changes in the chapter.

Here is what is different compared to the formalisation from the book:

```agda
open import Data.Empty using (⊥)
open import Data.Nat as Nat using (ℕ; zero; suc; _<_; s≤s; z≤n)
open import Data.Nat.Properties using (≤-pred)
open import Data.Unit using (⊤; tt)


length : Context → ℕ
length ∅        =  zero
length (Γ , _)  =  suc (length Γ)

lookup' : (Γ : Context) → (n : ℕ) → (n < length Γ) → Type
lookup' (_ , A) zero _         =  A
lookup' (Γ , _) (suc n) ≤-suc  =  lookup' Γ n (≤-pred ≤-suc)

count' : ∀ {Γ} → (n : ℕ) → (s : n < length Γ) → Γ ∋ lookup' Γ n s
count' {_ , _} zero _         =  Z
count' {Γ , A} (suc n) ≤-suc  =  S (count' n (≤-pred ≤-suc))

_≤?_ : ∀ (m n : ℕ) → Set
[...]

PR #514, opened 2020-09-11
DeBruijn: introduce a version of #_ that depends on no postulate

mdimjasevic:
Per the discussion in issue #513, this patch introduces a formalisation of variables based on de Bruijn indices that relies on no postulates. The accompanying text is updated accordingly.

### DIFF
--- a/src/plfa/part2/DeBruijn.lagda.md
+++ b/src/plfa/part2/DeBruijn.lagda.md
@@ -1376,7 +1376,7 @@ number of lines of code is as follows:
 
     Lambda                      216
     Properties                  235
-    DeBruijn                    275
+    DeBruijn                    276
 
 The relation between the two approaches approximates the
 golden ratio: extrinsically-typed terms
--- a/src/plfa/part2/DeBruijn.lagda.md
+++ b/src/plfa/part2/DeBruijn.lagda.md
@@ -454,7 +454,7 @@ We can then introduce a convenient abbreviation for variables:
 The type of function `#_` asks for clarification. The function takes
 an implicit argument `n<?length` that signals if there is evidence for
 `n` to be within the context's bounds. Both `True` and `_≤?_` are
-defined in Chapter [Decidable]({{ site.baseurl }}/Decidable/. The type
+defined in Chapter [Decidable]({{ site.baseurl }}/Decidable/). The type
 of `n<?length` guards against invoking `#_` on an `n` that is out of
 context bounds. Finally, in the return type `n<?length` is converted
 to a witness that `n` is within the bounds.
--- a/src/plfa/part2/DeBruijn.lagda.md
+++ b/src/plfa/part2/DeBruijn.lagda.md
@@ -272,16 +272,6 @@ is a context with two variables in scope, where the outer
 bound one has type `` `ℕ ⇒ `ℕ ``, and the inner bound one has
 type `` `ℕ ``.
 
-We define a helper function that computes the length of a context,
-which will be useful later in making sure an index is within context
-bounds:
-
-```
-length : Context → ℕ
-length ∅        =  zero
-length (Γ , _)  =  suc (length Γ)
-```
-
 ### Variables and the lookup judgment
 
 Intrinsically-typed variables correspond to the lookup judgment.
@@ -423,11 +413,19 @@ The final term represents the Church numeral two.
 
 ### Abbreviating de Bruijn indices
 
+We define a helper function that computes the length of a context,
+which will be useful in making sure an index is within context bounds:
+```
+length : Context → ℕ
+length ∅        =  zero
+length (Γ , _)  =  suc (length Γ)
+```
+
 We can use a natural number to select a type from a context:
 ```
-lookup : (Γ : Context) → (n : ℕ) → (p : n < length Γ) → Type
-lookup (_ , A) zero    (s≤s z≤n)  =  A
-lookup (Γ , _) (suc n) (s≤s p)    =  lookup Γ n p
+lookup : {Γ : Context} → {n : ℕ} → (p : n < length Γ) → Type
+lookup {(_ , A)} {zero}    (s≤s z≤n)  =  A
+lookup {(Γ , _)} {(suc n)} (s≤s p)    =  lookup p
 ```
 The function takes a proof `p` that the natural number is within the
 context's bounds. This guarantees we will never try to select a type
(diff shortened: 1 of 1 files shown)


src/plfa/part2/DeBruijn.lagda.md
```diff
@@ -271,6 +272,16 @@ is a context with two variables in scope, where the outer
 bound one has type `` `ℕ ⇒ `ℕ ``, and the inner bound one has
 type `` `ℕ ``.
 
+We define a helper function that computes the length of a context,
+which will be useful later in making sure an index is within context
+bounds:
```
wenkokke: If it isn't going to be used until later, than perhaps we shouldn't define it until it is useful.

src/plfa/part2/DeBruijn.lagda.md
```diff
@@ -414,33 +425,42 @@ The final term represents the Church numeral two.
...
+#_ n {p}  =  ` count n (toWitness p)
+```
+The type of function `#_` asks for clarification. The function takes
+an implicit argument `p` that signals if there is evidence for `n`
+to be within the context's bounds. If `n` is within the bounds, `p`'s
+type `True (suc n ≤? length Γ)` will compute to `⊤`, the familiar unit
+type with only one value. If `n` is not within the bounds, the type
+will compute to `⊥`, which is an inhabited type and the call to `#_`
+for such an `n` will fail to type-check. Finally, in the return type
+`p` is converted to a witness that `n` is within the bounds.
```
wenkokke: The `True` function is defined in the Decidable chapter, so instead of duplicating the explanation, it's probably better to point the reader at the relevant chapter.

src/plfa/part2/DeBruijn.lagda.md
```diff
@@ -414,33 +425,42 @@ The final term represents the Church numeral two.
...
 We can use a natural number to select a type from a context:
 ```
-lookup : Context → ℕ → Type
-lookup (Γ , A) zero     =  A
-lookup (Γ , _) (suc n)  =  lookup Γ n
-lookup ∅       _        =  ⊥-elim impossible
-  where postulate impossible : ⊥
+lookup : (Γ : Context) → (n : ℕ) → (p : n < length Γ) → Type
+lookup (_ , A) zero    (s≤s z≤n)  =  A
+lookup (Γ , _) (suc n) (s≤s p)    =  lookup Γ n p
```
wenkokke: I may be worth emphasizing that a proof `n < length Γ` contains enough information to be used in place of the index `n`, and so you could let `n` be an *implicit* argument, and instead define `lookup` and `count` by induction on the proof `p`.

src/plfa/part2/DeBruijn.lagda.md
```diff
@@ -414,33 +425,42 @@ The final term represents the Church numeral two.
...
+     --------------------------------
+   → Γ ⊢ lookup Γ n (toWitness p)
+#_ n {p}  =  ` count n (toWitness p)
+```
+The type of function `#_` asks for clarification. The function takes
+an implicit argument `p` that signals if there is evidence for `n`
+to be within the context's bounds. If `n` is within the bounds, `p`'s
+type `True (suc n ≤? length Γ)` will compute to `⊤`, the familiar unit
+type with only one value. If `n` is not within the bounds, the type
+will compute to `⊥`, which is an inhabited type and the call to `#_`
```
mdimjasevic: This should have been "uninhabited" type.

src/plfa/part2/DeBruijn.lagda.md
mdimjasevic: Sure, moving it to just before the formalisation of `lookup`.

src/plfa/part2/DeBruijn.lagda.md
mdimjasevic: That is a good point! I'm updating the text with a pointer to Chapter Decidable.

src/plfa/part2/DeBruijn.lagda.md
mdimjasevic: That's sounds like a reasonable suggestion.

I've updated the corresponding functions to have `n` as an implicit argument.

wenkokke (approved):

wadler (changes requested):
There appears to be a serious typo: No closing paren!

defined in Chapter [Decidable]({{ site.baseurl }}/Decidable/. The type

mdimjasevic:
> There appears to be a serious typo: No closing paren!
> 
> defined in Chapter [Decidable]({{ site.baseurl }}/Decidable/. The type

You are right. I just added the missing parenthesis.

wenkokke:
Why did you close this? If you just push to the same branch, it'll automatically update the pull request.

mdimjasevic:
> Why did you close this? If you just push to the same branch, it'll automatically update the pull request.

My apologies, this was not my intent. I misinterpreted the user interface here in Github.

How do I mark @wadler 's request for changes as done?

wenkokke:
> > Why did you close this? If you just push to the same branch, it'll automatically update the pull request.
> 
> My apologies, this was not my intent. I misinterpreted the user interface here in Github.
> 
> How do I mark @wadler 's request for changes as done?

You can't, you'll have to wait for them to review your code again.

wadler (commented):
Nice!

454: "The type of function `#_` asks for clarification. The function" --> "Function `#_`
456: "that signals if there is evidence" --> "that provides evidence"
456/7: "Both `True` and `_≤?_` are
 defined in Chapter [Decidable]({{ site.baseurl }}/Decidable/)." -->
"Recall that `True`, `_≤?_`, and `toWitness` are
 defined in Chapter [Decidable]({{ site.baseurl }}/Decidable/),
[here], [here], and [here]"  [Give appropriate links to the exact definitions]

wenkokke:
Links with just the word "here" are discouraged for accessibility reasons, so perhaps it's better to just add the links to True, _≤?_, and toWitness instead?

mdimjasevic:
Dear @wadler and @wenkokke , I just updated the pull request with your feedback from today. Hopefully it does what you had in mind. I linked to sections in Chapter Decidable instead of to the definitions themselves. If there is a way to link to the code blocks providing the definitions, please let me know and I will happily make another update to the pull request.

wadler (changes requested):
Great! All that remains to do is to delete "The type of function `#_` asks for clarification." as per my previous message.

mdimjasevic:
> Great! All that remains to do is to delete "The type of function `#_` asks for clarification." as per my previous message.

My apologies, I misunderstood the comment the first time. I hope I got it right this time!

wadler (approved):
Great, thank you!

PR #515, opened 2020-09-13
DeBruijn: Adds a conjunction for easier reading

mdimjasevic:
In the chapter on intrinsically-typed variables in section on renaming, this pull request adds a conjunction for easier reading, i.e., parsing of a sentence. Without it, it took me several consecutive readings of the sentence to get its meaning. Needless to say, the sentence is fine just the way it is, but I'd argue it is much easier to read it when a `that` is added as in this patch.


wadler (commented):
The "that" is not necessary, and leaves two "that"s in a row, which I find clunky. Thanks for trying to make it easier to read, but I prefer the original.

PR #521, opened 2020-09-29
Close quotation mark

OlingCat:


wenkokke:
Thanks, @OlingCat!

OlingCat:
@wenkokke  Sorry, I just noticed this fix was wrong. It's from another place as a qoute. can you revert that?

wenkokke:
Ah, I merged over-eagerly. Thanks for spotting that! We should probably fix the quotation marks so in doesn't have nested double quotes.

wadler (commented):
Thanks! But isn't there an extra close quote at the end of the paragraph which needs to go?

PR #522, opened 2020-09-29
Revert "Close quotation mark"

wenkokke:
Reverts plfa/plfa.github.io#521


wadler:
It appears this section contains a long quote from elsewhere. If that is correct, the source of the quote must be given! As it stands, it reflects poor scholarly practice, bordering on plagiarism.

wenkokke:
It does, it's a quote from the Agda docs.

wenkokke:
Fixed.

PR #523, opened 2020-09-29
Definitons of functions in exercise 'primed' without postulates

mdimjasevic:
This patch is to the introductory chapter to lamba calculus. It proposes to use no postulates in the definitions of functions `ƛ′_⇒_`, `case′_[zero⇒_|suc_⇒_]` and `μ′_⇒_` in the exercise `primed`.

If this sounds reasonable to you, I will update all the references to these definitions throughout the book.

### DIFF
--- a/src/plfa/part2/Lambda.lagda.md
+++ b/src/plfa/part2/Lambda.lagda.md
@@ -53,12 +53,13 @@ four.
 ## Imports
 
 ```
-open import Data.Bool using (T; not)
+open import Data.Bool using (Bool; true; false; T; not)
 open import Data.Empty using (⊥; ⊥-elim)
 open import Data.List using (List; _∷_; [])
 open import Data.Nat using (ℕ; zero; suc)
 open import Data.Product using (∃-syntax; _×_)
 open import Data.String using (String; _≟_)
+open import Data.Unit using (tt)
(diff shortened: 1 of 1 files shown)


wadler:
Looks like a neat idea, thanks! But perhaps there needs to be more explanation in the text? Wen, what do you think?

wenkokke:
I can make some time to look into this, but it may take a couple of weeks.

mdimjasevic:
> Looks like a neat idea, thanks! But perhaps there needs to be more explanation in the text? Wen, what do you think?

I will be happy to add more explanation! What should I add?

mdimjasevic:
> Looks like a neat idea, thanks! But perhaps there needs to be more explanation in the text? Wen, what do you think?

@wadler , I updated the text by providing more explanation and by linking to Chapter Decidable. Does it look good now?

wenkokke:
@wadler I recall you wanted to leave the `primed` execise in, specifically to introduce readers to the idea that they *can* write partial functions by postulating impossible, but that they should only do this for functions they know will only ever be used statically with correct inputs... i.e., when they feel using the trick with decidable is too much work?

wadler (approved):
Fantastic! Thank you.

PR #526, opened 2020-10-05
De Bruijn: fix a line-break in the middle of a markup

mdimjasevic:
Currently, in the de Bruijn chapter there is a line-break in the middle of a markup, which results in a weird rendering, as below. This patch fixes that.

markup-line-break


wadler (commented):
Thank you!

PR #528, opened 2020-10-10
De Bruijn: makes type A a parameter to the Steps data type

mdimjasevic:
In the chapter on de Bruijn indices, this patch changes the `Steps` data type such that type `A` is its parameter instead of index. This makes it unnecessary to quantify over `A` in the data constructor `steps`, even more so because both `L` and `N` are of type `∅ ⊢ A`. In other words, no need to unify two `A`s because they are the same anyway.


wenkokke:
Yep, that looks better!

wadler (commented):
Much better. Thank you!

PR #530, opened 2020-10-14
More: fixes an informal definition of swap×

mdimjasevic:
A disadvantage of informal presentation is that it doesn't get type-checked by Agda, and this allowed the definition of `swap×` to refer to undefined projection functions. This patch fixes the error in the definition.


wadler (approved):
Thank you!

PR #531, opened 2020-10-14
More: fixes the syntax of substitution in rule β-case×

mdimjasevic:
In the chapter with additional constructors for a lambda calculus, this patch fixes the syntax of substitution in the rule β-case×. If there was no whitespace between the consecutive square brackets, this rule would not type-check if formalised via definitions from the introductory chapter on lambda calculus.


wadler (changes requested):
This is not an error! If you search for _[_][_] you will find that simultaneous substitution of two variables is different from consecutive substitution of one variable and then another variable.

PR #532, opened 2020-10-14
More: fixes the syntax of substitution in a wrong translation of case×

mdimjasevic:
In the chapter on additional constructs of a lambda calculus, this patch fixes the syntax of substitution in an informal intentionally wrong translation of `case×` by adding a whitespace between consecutive square brackets. It also fixes the names of projection functions so they refer to the development earlier in the chapter.


wadler (changes requested):
This is not an error! See comment on previous pull request.

wadler (approved):
Thank you! This is analogous to the error with proj you caught earlier.

PR #533, opened 2020-10-20
More: Fixes the Projection Function Used in the Informal Definition of from×⊤

mdimjasevic:
In the chapter on more constructs in a lambda calculus, this patch fixes a reference to a projection function such that it refers to the one defined in the chapter rather than to a non-imported Agda projection function.


wenkokke:
Thanks for all those fixes, @mdimjasevic !

mdimjasevic:
You are more than welcome, @wenkokke ! I am happy to see you've been making changes yourself all over the place! I have a few more immediate small patches to submit based on your changes.

PR #544, opened 2020-11-05
Bisimulation: Fixes a Verb Form

mdimjasevic:
In Chapter Bisimulation, this tiny patch fixes a verb form. The current form, when expanded, would read "let us us".


wadler (approved):
Good catch! Thank you.

PR #546, opened 2020-11-25
For pulls

jphmrst2:
My undergrads needed much more detail on getting Agda etc. installed, and I ended up adding considerably more detail to the README/GettingStarted page.
 - Bringing in more specifically only what is required here from pages just linked in the original.
 - Adding more details about installing prerequisites.
 - Adding details specific to Mac/Windows/Linux for various steps.

I've moved the instructions for generating standalone Agda binaries to an "optional" appendix.  These seem to not be needed for PLFA, but I'll confess that I did not closely read the sections that are well beyond what my class will be able to cover.

There is also an appendix intended for system admins installing Agda (in particular the Agda standard library) on shared machines/network drives.  The obvious thing is to make the stdlib read-only, but then Agda cannot save the .agdai files the first time it loads the stdlib.  So there is a great big .agda file quoted there that simply forces tham all to be compiled.  The idea is that superuser sysadmin runs this file, and then makes the .agdai files available for sharing as well.  Arguably the .agda file should be a separate link of its own, but I'll leave that to you good people.

Lastly, there is a separate Troubleshooting page collecting every issue that I or my students could figure out.  This pull request does not link that file from any other .md page.  In my local setup, it's an extra indented bullet point in index.md, right below the link to README/Getting started.  But if it's something you want to include, I'll leave the linking to it to you.

Hope this helps --- thanks for all the work you've put into this great book.

-John

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ download that version, and then ask Stack to install it for us.
     ```
 
     The last step may give you notices about being in "'detached HEAD'
-    state" --- it's fine.  The success message you hope to see is
+    state" — it's fine.  The success message you hope to see is
     `Note: switching to 'v2.6.0.1'`.
 
     Alternatively, you can download a ZIP archive of that version from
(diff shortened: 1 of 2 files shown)


jphmrst2:
I noticed after submitting the pull request that much of this is now redundant with your README --- sorry, I thought I had taken all of the updates from origin, but clearly I missed some!  

You may still be interested in the sysadmins' appendix, and in the Troubleshooting page.

wenkokke:
Thanks for the pull request, but as you mentioned, I'd already merged a lot of this advice.

wadler:
Thanks for the effort you put into this, John!

Issue #547, opened 2020-12-01
Part 1 - Equality chapter - Proving transitivity using ≡-Reasoning is circular


crisoagf:
Hi! I am having a lot fun reading PLFA and following with the exercises. It's great!

However, while I was reading the Equality chapter something rubbed me the wrong way. ≡-Reasoning is defined, and then the first thing that is done is using ≡-Reasoning to prove transitivity of ≡.
However, ≡-Reasoning requires transitivity, so this could never be the actual definition of `trans`, and in no place is this even mentioned.
Would it be reasonable to mention this (or add an exercise to think about this)?

wadler:
Good point! Yes, please make a pull request.

crisoagf:
On it.

crisoagf:
🎉🎉🎉

PR #548, opened 2020-12-03
Create an exercise to reflect the fact that proving trans using ≡-Reasoning is circular

crisoagf:
This a PR proposal for addressing #547.

I added `-- Your code goes here` because all exercises seem to include this, even pen and paper exercises.

Let me know if this is not what you want.


wadler (approved):
Thank you! Neatly done.

PR #551, opened 2020-12-30
Added hint to "_t" exercise in Bisimulation chapter.

capn-freako:
After an e-mail exchange with Dr. Wadler, re: the "_t" exercise in the _Bisimulation_ chapter of Part 2 of PLFA, he asked that I add a hint relating to our discussion.
This pull request contains that hint.


wadler (changes requested):
Thanks! I suggest a few changes.
1. After "**Hint**:" add: Because we focus on only a few language constructs, ".
2. In `Incomplete pattern matching for _†.` remove period at end.
3. Change `|~` to `⊢'`. (Where ' is prime, typed \'.)
4. Change "You'll need to write a mapping function that takes terms of this new type into equivalents in the original type: `⊢`, in order to use the `~` relation, as defined, in your proofs." to "Map `_†` should take terms in the shrunken type `⊢'` into the full type `⊢`."

capn-freako:
Requested changes made and pushed.

**Note**: I've assumed you meant "Map _†**'** should..." in 4.

capn-freako:
I'm realizing that I don't understand 4 above.
It seems like `map _†` takes a list of terms of type: `Γ ⊢ A`, to a list of _translated_ terms, also of type: `Γ ⊢ A`.

capn-freako:
After this reply from Dr. Wadler, in a _plfa-interest_ list thread:

> You are making things much harder than they need to be. Instead of

```agda
_′† : ∀ {Γ A}
  → Γ ⊢' A
    ------
  → Γ ⊢' A
```

> the signature should be

```agda
_† : ∀ {Γ A}
  → Γ ⊢' A
    ------
  → Γ ⊢ A
```

> (No prime on function and its result.) Go well. -- P

I've realized my mistake and pushed a correction to the added hint, which removes the (incorrect) prime on _†.

capn-freako:
Converted to draft mode upon realizing that the change to the type signature of the _† function may have unforeseen extensive consequences to the rest of the pre-existing code (i.e. - the code from the book).
Need to discuss this w/ Dr. Wadler.

It may be best to keep the _† exercise as an adjunct work, rather than trying to integrate it into the rest of the code base.
After all, the _† function is never used anywhere outside that exercise.

wadler (changes requested):
No, you've still missed the point. Here 'map' was just a synonym for function. Your misunderstanding shows that perhaps it is a poor synonym. So how about this? "Function `_†'` should take terms in the shrunken type `⊢'` into the full type `⊢`."

capn-freako:
I've gotten things working and updated the hint accordingly.
Please, see this file:

https://github.com/wadler/plfa-answers/blob/master/3rdParty/capn-freako/part2/Bisimulation.agda

for a complete solution that is consistent with the hint provided in this PR.

Thanks,
-db

wadler (commented):
Thanks for the prod. I've now revised the hint for the exercise along the lines you suggested (but with streamlined wording).

PR #553, opened 2021-01-08
fix README.md link to EPUB e-book

yeradis:
Simple update to fix a broken link to the EPUB e-book

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,7 +325,7 @@ If you want Stack to use your system installation of GHC, follow the instruction
 
 <!-- Links -->
 
-[epub]: https://plfa.github.io/out/epub/plfa.epub
+[epub]: https://plfa.github.io/plfa.epub
 [plfa]: http://plfa.inf.ed.ac.uk
 [plfa-dev]: https://github.com/plfa/plfa.github.io/archive/dev.zip
 [plfa-status]: https://travis-ci.org/plfa/plfa.github.io.svg?branch=dev


wadler (changes requested):
Thanks! The old link is clearly wrong. There are also problems with the new link, for instance no chapter headings appear in the epub. I have an older version without these errors. Wen, any idea what went wrong?

yeradis:
@wadler Hi

Don't really know what went wrong, arrived here looking for an offline version of the ebook and then found the link was broken

So, first time here,

But just downloaded the repo to build everything from scratch including the epub version of the ebook to see if there are some warning or errors

The correct version you have is available at some place so i can compare? 

Regards

wadler:
Wen should have an up-to-date copy of the correct ebook for comparison. Wen, can you supply this to Yeradis?

yeradis:
Hello again

I tried to look into this in multiple ways but got stuck cause i'm not able to run the full process on MacOS BigSur,
something have change on this OS version making compile process to fail, this seems to be fixed on recent GHC versions but not on the version that this projects need 
as a reference https://gitlab.haskell.org/ghc/ghc/-/issues/19037

Sorry but with my current knowledge about the toolchain i can not go deeper

Regards

wenkokke:
Unfortunately, EPUB generation is broken at the moment :( If someone has the expertise with Pandoc to help us fix it that would be greatly appreciated!

PR #558, opened 2021-02-11
Some Typo fixes.

capn-freako:
These are just some typo fixes I made, while making my way through the end of Part 2 and the _Denotational_ chapter.


wadler (changes requested):
This hint is out of date. I've already updated the text with an improved hint, based on discussion with David.

wadler (changes requested):
See comment on previous change.

wadler (changes requested):
See previous comment.

wadler (changes requested):
See previous comment.

wadler (commented):
Jeremy needs to review these.

wadler:
Most of these changes are to the dagger exercise in Bisimulation, which I already fixed after a discussion with David, so they are out of date. They should be deleted from the request. The other changes are to denotational and should not be bundled with these; they need to be approved by Jeremy, the principal author of that chapter.

jsiek:
I’ll take a look after the un-bundling.

Cheers,
Jeremy

capn-freako:
Please pardon my nascent Git-fu here, but I:

1. Can't seem to link any of your comments to a particular change.
2. Don't see any changes to _Bisimulation_ in the pull request. (I see two files changed: _Inference_ and _Denotational_.)
3. Am not sure how to split this pull request, now that it's been submitted.

Thanks,
-db

jsiek:
The suggested changes to the Denotational chapter look good.

src/plfa/part2/Inference.lagda.md
```diff
@@ -188,15 +188,15 @@ We can extract the grammar for terms from the above:
     L⁺, M⁺, N⁺ ::=                      terms with synthesized type
       x                                   variable
       L⁺ · M⁻                             application
-      M⁻ ↓ A                              switch to inherited
+      M⁻ ↓ A                              switch to synthesized
```
jsiek, line 191: It looks to me like "switch to inherited" was correct, so no change needed.

src/plfa/part2/Inference.lagda.md
```diff
@@ -188,15 +188,15 @@ We can extract the grammar for terms from the above:
...
 
     L⁻, M⁻, N⁻ ::=                      terms with inherited type
       ƛ x ⇒ N⁻                            abstraction
       `zero                               zero
       `suc M⁻                             successor
       case L⁺ [zero⇒ M⁻ |suc x ⇒ N⁻ ]     case
       μ x ⇒ N⁻                            fixpoint
-      M⁺ ↑                                switch to synthesized
+      M⁺ ↑                                switch to inherited
```
jsiek, line 199: It looks to me like "switch to synthesized" was correct, so no change needed.

capn-freako:
But, isn't that inconsistent with both the text above:

Hence, we need a way to treat a synthesized term as if it is inherited. We
> introduce a new term form, M ↑ for this purpose.


and the definition:

  _↓_ <https://plfa.github.io/Inference/#12229>
: Term⁻ <https://plfa.github.io/Inference/#12104> → Type
<https://plfa.github.io/Inference/#11725> → Term⁺
<https://plfa.github.io/Inference/#12087>


?

jsiek:
My understanding is that the text above is written from the perspective
of going from inside to out… you’ve got a synthesized term M
and want to use it in a context that in inherited, so you write M ↑.

The grammar comments are written from the perspective of
going from outside to inside. If you’re in an inherited context
and want to fill it in with a synthesized term M, you write M ↑.

Perhaps the best fix is to adjust the grammar comments or the text
to take the same perspective.

Cheers,
Jeremy

capn-freako:
Okay, but in either case below, the expression: M ↑, is taking a synthesized term and making it an inherited term; correct?

I think the comments in the code currently have that reversed.

-db

jsiek:
> On Feb 24, 2021, at 11:03 AM, David Banas <notifications@github.com> wrote:
> 
> 
> Okay, but in either case below, the expression: M ↑, is taking a synthesized term and making it an inherited term; correct?
> 

Yes

> I think the comments in the code currently have that reversed.
> 

What I was trying to say below is that the comment

"switch to synthesized"

should be interpreted to mean

“switch from an inherited context to a synthesized context”

and not

“switch an inherited term into a synthesized term”

Cheers,
Jeremy


> -db
> 
>

wenkokke:
Closed due to inactivity.

Perhaps submitting some of these changes as separate PRs would make it easier to merge them quickly.

PR #562, opened 2021-03-09
Fix typos and add a missing Unicode symbol

lacrosse:
For some reason my agda-mode produces `⌞` and `⌟` for  `\cll` and `\clr`, respectively, so I fixed those shortcuts. Let me know if this should be reverted.


wadler (approved):
Thank you for all the fixes!

PR #563, opened 2021-06-02
fix typo in equality

priyasiddharth:
I believe the type of the second  argument to `_≡_` should be `A`.


priyasiddharth:
@wadler please take a look

wadler (changes requested):
I think the current wording is fine. (Arguably, A -> Set is correct and the current wording is not, but I think the current wording is less confusing.)

wadler:
Please see review.

PR #564, opened 2021-06-09
Link to `cask-fonts` cask for OSX.

jasonhemann:
Homebrew doesn't come out-of-the-box with mononoki available; this gives the user a hint as to how to install


wadler (approved):
Thank you!

PR #568, opened 2021-06-25
fixed indentation for proper code displays in Substitution.lagda.md

Altariarite:
These are code displays within a list, which must be indented by eight spaces. Originally they are indented with 4 spaces, so the code blocks are not rendered correctly on the website. That's because when writing code blocks in lists, markdown requires 8 spaces as a minimum(https://meta.stackexchange.com/questions/3792/how-to-nest-code-within-a-list-using-markdown).

The pull request should fix this issue by changing the indentation to 8 spaces.

### DIFF
--- a/src/plfa/part2/Substitution.lagda.md
+++ b/src/plfa/part2/Substitution.lagda.md
@@ -807,11 +807,11 @@ We proceed by induction on the term `M`.
 
 * If `M = ƛ N`, we first use the induction hypothesis to show that
 
-     ƛ ⟪ exts σ₂ ⟫ (⟪ exts σ₁ ⟫ N) ≡ ƛ ⟪ exts σ₁ ⨟ exts σ₂ ⟫ N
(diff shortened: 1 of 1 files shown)


wadler (changes requested):
As far as I know, indenting code blocks by 4 spaces is correct. (I just checked, and that is what happens in the source for Connectives.)

Altariarite:
4 spaces are normally correct, but when mixed with the markdown list notation then 8 spaces are required as a minimum, which is the case here. 

Substitution.lagda.md actually is already indented by 8 spaces on line 728 and 736 (around 'We prove this equation by cases on x'). This indentation leads to proper display on the website:

image

However,  line 810 and 814 (the two lines I changed) were originally indented by 4 spaces, and the code is not in codeblocks as expected:
image

wadler (changes requested):
Yuanting explained that the key point is these are *code displays within a list*, which must be indented by eight spaces. Thanks! Please update the introductory comment to mention this (it currently mentions displays but not lists, save for the link name), and then I will approve.

Altariarite:
I've edited the comment. I've also included another fix at 736 where the codeblocks need to be indented deeper because they are in a sublist.

wadler (approved):
Thanks!

PR #569, opened 2021-07-02
changed codeblocks to displays in Fonts.lagda.md

Altariarite:
For compatibility with LaTex, code blocks in Fonts.lagda.md need to be changed into displays for the following reasons:

1. There's a blank line at the end of the comment, and Agda doesn't know how to handle that: https://github.com/agda/agda/issues/5398
2. If I delete the blank line, the pdf could compile, but all comments will be squeezed onto the same line. (Possibly a bug in Agda)
image


wadler (approved):
Thank you!

PR #570, opened 2021-07-16
Include home brewed font to make the code monospaced

Altariarite:
This pull request contains a home-brewed font that's a mix of DejaVu Sans Mono, mononoki, Symbola and Droid Sans Mono.

It should make all vertical bars line up under Fonts.lagda.md: "Preferably, all vertical bars should line up."

Currently, the font includes emoji from Symbola that will be rendered in black and white. I could make a trimmed version so that the browser will use the system Emoji instead, if that's better.


wadler:
Waiting for Wen to approve.

Omitting emoji so that they can be pulled from another font makes sense. It's not required for emoji to be monospace.

PR #571, opened 2021-07-26
Fix typo replacing "monus" with minus

kisonecat:


wadler (changes requested):
No, "monus" is the correct name.

PR #573, opened 2021-07-29
Change some instructions in the README.md

L-TChen:
- Only clone PLFA shallowly. I am aware that some users may have rather limited Internet access.

- Update the version of the standard library checked out for PLFA. The instruction in `README.md` checks out `v1.3` instead of `v1.6` specified by the git submodule.


L-TChen:
`fix-whitespace` is available on Stackage, so there is no need to download the source files to install.

wadler:
Thank you! I'm going to leave Wen to review this pull request.

wenkokke:
Thanks, @L-TChen !

PR #590, opened 2021-08-31
Fix #589

### LINKED ISSUE #589: part1/Quantifiers: exercise requires proj₁, which can't be imported nor defined due to conflicts
[Exercise Bin-isomorphism](https://plfa.github.io/20.07/Quantifiers/#Bin-isomorphism) asks the reader to give evidence for:
```agda
proj₁≡→Can≡ : {cb cb′ : ∃[ b ] Can b} → proj₁ cb ≡ proj₁ cb′ → cb ≡ cb′
```
But, when the user attempts the exercise he is faced with an error:
```agda
postulate
  Bin : Set
  Can : Bin → Set

proj₁≡→Can≡ : {cb cb′ : ∃[ b ] Can b} → proj₁ cb ≡ proj₁ cb′ → cb ≡ cb′
proj₁≡→Can≡ = {!!}
```
```
(Σ Bin (λ b → Can b)) !=< (Data.Product.Σ _A_156 _B_157)
when checking that the expression cb has type
Data.Product.Σ _A_156 _B_157
```

This is due to the fact that the current file imports `proj₁` from `Data.Product`, meaning that `proj₁` will work on the Σ type from the stdlib, but not on the Σ type defined in this chapter.

I'm not sure how it would be best to handle this problem, but here are a few observations:
1) Although the module imports `proj₁` and `proj₂`, it doesn't actually use them, hence we could remove those imports and ask the reader to define a suitable `proj₁` for the Σ in this chapter; but, the problem would arise again if the reader himself imports `proj₁` to solve previous exercises.
2) The exercise could be reframed in terms of `Σ′` which has been defined as a record, with a `proj₁′` function available already; but this way we'd lose the nice syntax which has been defined for `Σ`.
3) We could explicitly mention the conflict, and ask the reader to define a `proj₁′` function for the Σ type in this chapter (this should not create conficts with `Σ′`, as its projection is namespaced as `Σ′.proj₁′`).

lorenzo-lipparini:
This pull request addresses issue #589, implementing the following changes in part1/Quantifiers:
1) `proj₁` and `proj₂` are not imported anymore from the standard library;
2) The definitions of `proj₁` and `proj₂` for the Σ type are added as an exercise for the reader, mentioning that they will collide with the ones standard library (in case they are deliberately imported);
3) Some minor style fixes in the statement of exercise Bin-isomorphism.

A few points which may be worth discussing:
1) I could contribute a solution to the Bin-isomorphism exercise, if I get access to the solutions repository.
2) The projection functions have been left as an exercise; I think this has the benefit of condensing the type definitions and the remark about the collisions in a few lines, so that they are easier to see in case an error occurs; also, it gives a showcase for `Σ-syntax`, which is otherwise unused in code; still, that task might be too trivial to be called an exercise.
3) The remark below the exercise includes a suggestion to use a renaming import; we could instead do the renaming import upfront, as it is done already for the constructor `_,_`.

### DIFF
--- a/src/plfa/part1/Quantifiers.lagda.md
+++ b/src/plfa/part1/Quantifiers.lagda.md
@@ -19,7 +19,7 @@ import Relation.Binary.PropositionalEquality as Eq
 open Eq using (_≡_; refl)
 open import Data.Nat using (ℕ; zero; suc; _+_; _*_)
 open import Relation.Nullary using (¬_)
-open import Data.Product using (_×_; proj₁; proj₂) renaming (_,_ to ⟨_,_⟩)
+open import Data.Product using (_×_) renaming (_,_ to ⟨_,_⟩)
 open import Data.Sum using (_⊎_; inj₁; inj₂)
 open import plfa.part1.Isomorphism using (_≃_; extensionality)
 ```
@@ -175,6 +175,27 @@ corresponds to the term
     ⟨ M , N ⟩
 
 where `M` is a term of type `A` and `N` is a term of type `B M`.
+
+Note that the record definition gives us the projection functions
+`Σ′.proj₁′` and `Σ′.proj₂′` for free; if we adopt the inductive
+type version, those have to be defined separately.
+
+#### Exercise `proj₁` and `proj₂` (recommended)
+
+Define the projection functions for existentials:
+```
+postulate
+  proj₁ : ∀ {A : Set} {B : A → Set} → (pair : Σ[ x ∈ A ] B x) → A
+  proj₂ : ∀ {A : Set} {B : A → Set} → (pair : Σ[ x ∈ A ] B x) → B (proj₁ pair)
+```
+These definitions will collide with those in the standard library,
+which is why `proj₁` and `proj₂` from `Data.Product` are not imported
+in this chapter; in case you need them to complete some exercises,
+you can use a renaming import.
+
+```
+-- Your code goes here
+```
 
 Products arise as a special case of existentials, where the second
 component does not depend on a variable drawn from the first
@@ -453,6 +474,7 @@ and asks you to define the following functions and predicates:
 
 And to establish the following properties:
 
+    ---------------
     from (to n) ≡ n
 
     ----------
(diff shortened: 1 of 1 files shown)


wadler:
Three of the exercises use products: ∀-distrib-×, ∀-×, and ∃×-implies-×∃. Before removing proj₁ and proj₂, please confirm that those exercises can be solved easily using only the constructor for products, possibly in pattern position. (I believe that is the case.)

Please remove the text at lines 191-194:
"These definitions will collide with those in the standard library,
which is why `proj₁` and `proj₂` from `Data.Product` are not imported
in this chapter; in case you need them to complete some exercises,
you can use a renaming import."
Assuming we confirm the projections are not needed for the exercises, I think this is as likely to be confusing as helpful.

Please add proj₁ and proj₂ for dependent product to the section "Standard Library".

lorenzo-lipparini:
@wadler of the three exercises you indicated, `∀-distrib-×` turns out to be the only one that is actually affected: using`proj₁` and `proj₂` the `to` function could be implemented as a one-liner, but without them some form of auxiliary pattern matching is required:
```agda
∀-distrib-× : ∀ {A : Set} {B C : A → Set} →
  (∀ (x : A) → B x × C x) ≃ (∀ (x : A) → B x) × (∀ (x : A) → C x)
∀-distrib-× {A} {B} {C} =
  record
    { to = to
    ; from = λ{ ⟨ ∀x→B , ∀x→C ⟩ x → ⟨ ∀x→B x , ∀x→C x ⟩ }
    ; from∘to = λ x → refl
    ; to∘from = λ y → refl
    }
  where
  to : (∀ (x : A) → B x × C x) → (∀ (x : A) → B x) × (∀ (x : A) → C x)
  to ∀x→B×C = ⟨ ∀x→B , ∀x→C ⟩
    where
    ∀x→B : ∀ (x : A) → B x
    ∀x→B x with ∀x→B×C x
    ... | ⟨ B-x , _ ⟩ = B-x

    ∀x→C : ∀ (x : A) → C x
    ∀x→C x with ∀x→B×C x
    ... | ⟨ _ , C-x ⟩ = C-x
```

@wadler I'll wait for your feedback to proceed with the rest of the changes.

wenkokke:
Closing due to inactivity.

PR #598, opened 2021-09-16
Remove the unused Eq in Relations

favonia:


favonia:
Hmm. Maybe `Eq` was used in the intended solutions? I did not check those.

wenkokke:
We likely did this to keep the line lengths down.

favonia:
> We likely did this to keep the line lengths down.

How about the following, then?
```agda
open import Relation.Binary.PropositionalEquality
   using (_≡_; refl; cong)
```

wenkokke:
That would work, but why would it be preferable?

favonia:
> That would work, but why would it be preferable?

Because `Eq` is unused. However, I do not have a strong opinion on this---this PR was meant to be a small style clean-up only.

PR #605, opened 2021-10-10
Properties: may fix a typo?

alissa-tung:
Since `Γ` is a context and `Γ , y ⦂ B` is an extended context, plus `x` is a variable.  The constructors `S` and `Z` indicate the location of a variable in the context...  I am not sure whether if it is more or less correct to use the word "context" here.


wadler (approved):
Thank you!

PR #606, opened 2021-10-14
Injectivity of context lookup is functionality

favonia:
Personally I feel `lookup` is more like a function from variables to types, and "injectivity" is functionality from this point of view. I forgot to say on the mailing list that this change was originally suggested by my (anonymous) student.


wadler (approved):
Thanks!

PR #610, opened 2021-10-21
No need to use postulate in the Inference chapter

favonia:


wadler (changes requested):
Nice catch! Chapter Lambda was updated to use proof-by-reflection, but the same update was not done here. But it would be even better to drop _≠_ completely, and instead reuse the smart constructor S' from that chapter.

favonia:
> Nice catch! Chapter Lambda was updated to use proof-by-reflection, but the same update was not done here. But it would be even better to drop _≠_ completely, and instead reuse the smart constructor S' from that chapter.

Did the new commit address the issue?

wenkokke:
Yep, thanks @favonia!

PR #619, opened 2021-11-21
Add a line pointing out the Y combinator in Untyped

### LINKED ISSUE #618: Should the Y combinator be pointed out in Untyped?
Should it be pointed out that the defined fixpoint in Untyped is well-known as the Y combinator, discovered by Haskell Curry? I learned about this the first time as the Y combinator. Such information might be helpful to learners but is now missing.

ShreckYe:
Closes #618

### DIFF
--- a/src/plfa/part2/Untyped.lagda.md
+++ b/src/plfa/part2/Untyped.lagda.md
@@ -721,6 +721,8 @@ This works because:
     ≡
       f · (μ f)
 
+This fixpoint definition is well-known as [the Y combinator](https://en.wikipedia.org/wiki/Fixed-point_combinator#Fixed-point_combinators_in_lambda_calculus), discovered by Haskell Curry.
+
 With de Bruijn indices, we have the following:
 ```
 μ_ : ∀ {Γ} → (Γ , ★ ⊢ ★) → (Γ ⊢ ★)


src/plfa/part2/Untyped.lagda.md
```diff
@@ -721,6 +721,8 @@ This works because:
     ≡
       f · (μ f)
 
+This fixpoint definition is well-known as [the Y combinator](https://en.wikipedia.org/wiki/Fixed-point_combinator#Fixed-point_combinators_in_lambda_calculus), discovered by Haskell Curry.
```
wenkokke, line 724: We probably shouldn't be referring to Wikipedia or saying Curry invented it without a source.

ShreckYe: OK. Shall I edit this or will you do it?

ShreckYe: OK tried tracing its source but I couldn't get access to the content of the second citation in the Wikipedia page, "Henk Barendregt (1985). The Lambda Calculus – Its Syntax and Semantics. Studies in Logic and the Foundations of Mathematics. 103. Amsterdam: North-Holland. ISBN 0444867481.".

wenkokke:
From _Combinatory Logic_ by Curry, Feys, and Craig:

> The  paradoxical  combinator  appeared explicitly in Rosenbloom [EML] (as Theta on pp. 130-131).  However. the combinator was implicit  in [IFL]; and the idea that the Russell  paradox  (and some others) could be formlilated in combinatory terms goes back  to the early days of  our subject (cf. [FPF], p. 373, which  was  essentially  con- tained  in  Curry’s letter of  December,  1929, to Hilbert, and Church [SPF.I], p. 347). For  a  similar  idea  see  Fitch  [SRR].

Excerpts from the bibliography:
> - [EML] ROSENBLOOM, P. C. The elements of  mathematical logic.  New  York,  1950.
> - [FPF] CURRY, H. B. First properties of  functionality in combinatory logic. Tōhoku Math. J. 41: 371-401 (1936).
> - [IFL] KLEENE, S. C.  and ROSSER, J. B. The inconsistency of certain formal logics. Ann. of Math. (2) 36 : 630- 636  (1935).
> - [SPF.I]  CHURCH, A. A  set of  postulates  for the foundation of  logic. Ann. of  Math.  (2) 33  346-366  (1932).
> - [SRR] FITCH, F.  R. Self-referential relations.  Actes  du XIème Congrès International  de Philosophie 14 : 121-127  (1953).

May I add that this work does not in any meaningful manner sort its bibliography.

src/plfa/part2/Untyped.lagda.md
wadler: When I said it should include a citation, I meant a citation to a book rather than a link to Wikipedia. (As Wen notes, links to Wikipedia are not proper academic citations.) It should be to the first person to name the combinator Y, which is not Barendregt.

I tried to track down the original citation just now, but failed. You can cite this paper, which is reasonably authorative, but doesn't actually say who first defined Y: http://www.users.waitrose.com/~hindley/SomePapers_PDFs/2006CarHin,HistlamRp.pdf. The footnote on the first page of the link gives the proper citation.

wenkokke:
Based on what Curry and Feys wrote, it seems incorrect to atribute the formulation of the Y combinator to Curry.

wenkokke:
> OK tried tracing its source but I couldn't get access to the content of the second citation in the Wikipedia page, "Henk Barendregt (1985). The Lambda Calculus.”

Barendregt merely gives the definition of Y and notes that Curry calls it the “paradoxical operator”.

wenkokke:
From [_History of Lambda-calculus and Combinatory Logic_](http://www.users.waitrose.com/~hindley/SomePapers_PDFs/2006CarHin,HistlamRp.pdf) by Felice Cardone and J. Roger Hindley:

> Incidentally, in the course of his doctoral work Turing gave the first published fixed-point combinator, [Turing, 1937b, term Θ]. This was seen as only having minor interest at that time, but in view of the later importance given to such combinators (see §8.1.2 below), we digress here to discuss them.
> 
> A fixed-point combinator is any closed term Y such that Yx converts to x(Yx). Turing’s was (λxy.y(xxy))(λxy.y(xxy)).
> 
> The next one to appear was in [Rosenbloom, 1950, pp.130–131, Exs. 3e, 5f]. It was λx. W(Bx)(W(Bx)), which is convertible to λx. (λy.x(yy))(λy.x(yy)). The latter has often been called Curry’s Y; in fact Curry gave no explicit fixed-point combinator before [Curry and Feys, 1958, §5G], but the accounts of Russell’s paradox in [Church, 1932, p.347] and [Curry, 1934a, §5] both mentioned (λy.N (yy))(λy.N (yy)), where N represented negation, and Curry’s use of the latter term dated back to a letter to Hilbert in 1929. Note that although fixed-point combinators are often used in λ-defining recursion, they are not really necessary for recursion on the natural numbers; none was used in the λ-representation of the recursive functions in [Kleene, 1936].

wadler:
I wrote to Hindley asking the proper citation for Y and he responded as follows: "I think the name Y was chosen by Curry; -- for no deeper reason than that other letters had already been chosen for other things.  His first reference to it is in Curry and Feys 1958  p.178  where he calls it the "paradoxical combinator"."  A citation to Cardone and Hindley (the passage mentioned by Wen above) seems best, as it lists all the major players (Turing, Rosenbaum, Curry).

wenkokke:
Closed due to inactivity.

PR #620, opened 2021-12-02
Fix a typo in BigStep

ShreckYe:
There seem to be no usages of the variable `c`. Judging from the context it should be `V` here.


jsiek (approved):
Looks good, thanks for catching that!

PR #623, opened 2021-12-12
Fix some typos and inconsistent symbols in Soundness

ShreckYe:
I also tried adding the missing backticks but then it doesn't work with Markdown syntax (see ShreckYe@b991fca66f799e640f8dae0302344bef2190feb9).


wenkokke:
I don't think that renaming the function to ``ext-`⊑`` is the solution? Could you explain why you've made the changes that you've made?

ShreckYe:
Sure. I did this because the subset symbol for `Env`s is `` `⊑ ``, and I thought the `⊑′` in `ext-⊑′` is `` `⊑ `` mistyped. It is still kind of inconsistent.

ShreckYe:
BTW, thanks for teaching me how to include a backtick inside an inline code block. 😂 I will also try to fix the missing backticks when I have time tomorrow then.

PR #628, opened 2021-12-30
Fix bulleted list for isomorphism record explanation

ashfordium:
Minor text change

The bulleted list for explaining the [isomorphism record](https://plfa.inf.ed.ac.uk/Isomorphism/#isomorphism) is broken for me. It looks like this:

image

This PR changes it to look like this:

image

I went with a numbered list since items are referred to as the 'third' and 'fourth'.


wadler (approved):
Thank you!

PR #630, opened 2022-03-08
remove paragraph that duplicates what's in the exercises

catern:
The presence of this paragraph here suggests that this approach should
be used for <-trans.  So that's what I did.  But then when I read
further I saw that this content/paragraph is duplicated in the ≤-iff-< and
<-trans-revisited exercises, and actually this approach is intended
for those exercises.  So I was a bit confused.

I don't know what the appropriate wording/ordering is here, but I
think just deleting this paragraph would be reasonable since the
content is duplicated in the ≤-iff-< and <-trans-revisited exercises.

### DIFF
--- a/src/plfa/part1/Relations.lagda.md
+++ b/src/plfa/part1/Relations.lagda.md
@@ -592,11 +592,6 @@ requires negation, as does the fact that the three cases in
 trichotomy are mutually exclusive, so those points are deferred to
 Chapter [Negation](/Negation/).
 
-It is straightforward to show that `suc m ≤ n` implies `m < n`,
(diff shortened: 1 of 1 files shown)


wadler (commented):
Thanks. Some readers skip the exercises, so I'd rather not delete this paragraph. Instead, I added clarification to the exercise on transitivity to explain one should give a direct proof.

wadler:
Thanks. Some readers skip the exercises, so I'd rather not delete this paragraph. Instead, I added clarification to the exercise on transitivity to explain one should give a direct proof.

PR #631, opened 2022-03-09
fix typo calling datatypes record types

catern:
Later on these are declared again as record types, which probably gave
rise to this error.


wadler (approved):
Thank you!

PR #632, opened 2022-03-11
lists: move postulate of foldr-++ up to the exercise, add missing "code goes here"

catern:
This follows the pattern set by the previous chapters, which always
postulate things proved by exercises and needed by later proofs, in
the exercise itself.  It's also slightly more convenient when doing the
exercise - one can just remove the "postulate" as a purely local
change, rather than having to scroll down and delete the postulate
from where it's used later.

Plus, the postulate of foldr-++ before took a _⊗_ : A → A → A, which
is insufficiently general for all the exercises, namely foldr-∷.

Just something I ran into, feel free to reject!

### DIFF
--- a/src/plfa/part1/Lists.lagda.md
+++ b/src/plfa/part1/Lists.lagda.md
@@ -353,6 +353,9 @@ reverse of the second appended to the reverse of the first:
 
     reverse (xs ++ ys) ≡ reverse ys ++ reverse xs
 
+```
+-- Your code goes here
+```
 
 #### Exercise `reverse-involutive` (recommended)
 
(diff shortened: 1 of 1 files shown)


wenkokke:
I’ll have a look at your changes, but I recall that in many places, formulating the type signature is meant as part of the exercise, and so we avoided postulating them on purpose. However, I am open to reviewing the places where we have done so.

wadler (approved):
Thanks!

wadler:
Thanks! I'm happy to add the missing "your code goes here" snippets, please submit as a separate pull request.

I don't want to change that one exercise to a postulate, as currently there are several exercises in the same form (without postulates). Separately, I've removed some redundant wording, thank you for drawing my attention to it.

PR #636, opened 2022-03-25
Turn off auto-fill-mode for the project

### LINKED ISSUE #516: .dir-locals.el doesn't work
The current .dir-locals.el breaks emacs (v28):

```elisp
dir-locals-read-from-dir: Wrong type argument: listp, turn-off-auto-fill
```

My understanding is that you can't call functions in .dir-locals.el. It's meant for variables.

KDr2:
Fix https://github.com/plfa/plfa.github.io/issues/516 correctly.


wadler:
@wenkokke needs to review.

wenkokke:
@KDr2: could you explain your fix?

KDr2:
> @KDr2: could you explain your fix?

I was reading and experimenting with the book in my emacs, and found `auto-fill-mode` was very annoying as others, then I find there was a `.dir-local.el` to disable that mode but it's disabled because it's buggy (actually in a bad format).

In the old file, it tried to disable auto-fill mode by calling a function directly, which is not allowed according to https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html. The doc says:

> The .dir-locals.el file should hold a specially-constructed list, which maps major mode names (symbols) to alists......Instead of a mode name, you can specify ‘nil’, which means that the alist applies to any mode;

So the new content `((nil . ((eval . (turn-off-auto-fill)))))` means, for all files under this directory, when it is opened, despite whatever major mode it will be applied to (`nil`), `(turn-off-auto-fill)` will be `eval`-ed.


Here is the doc for the `eval` file variable:
https://www.gnu.org/software/emacs/manual/html_node/emacs/Specifying-File-Variables.html
It says:

> `eval` evaluates the specified Lisp expression (the value returned by that expression is ignored).

PR #651, opened 2022-05-20
Fix #650

### LINKED ISSUE #650: Describe need for step-≡ constructor in equational reasoning syntax
https://github.com/plfa/plfa.github.io/blob/dev/src/plfa/part1/Induction.lagda.md

It isn't clear that `step-≡` within the following line is what imports what is being referred to as `_≡⟨_⟩_` in the rest of the text:

```
open Eq.≡-Reasoning using (begin_; _≡⟨⟩_; step-≡; _∎)
```

GetContented:
Adds explanation of where `_≡⟨_⟩_` comes from


wadler (commented):
I think the proposed new wording is as likely to confuse as to help readers not already familiar with syntax declarations.

Possibly and `_≡⟨_⟩_` (imported via `step-≡`), which are ..." would work, but I'm not so sure.

GetContented:
> I think the proposed new wording is as likely to confuse as to help readers not already familiar with syntax declarations.

Possibly. That wording would have helped me. All I really wanted to know was where that operation came from. Adding that it's defined as syntax would have been interesting and informative, too; I didn't know Agda had that facility.

> Possibly and _≡⟨_⟩_ (imported via step-≡), which are ..." would work, but I'm not so sure.

This would also work, IMO. I flirted with this exact wording several times.

We could add a note underneath, as seems to be the custom in the book; ie material is introduced, then it is explained:

Note: `_≡⟨_⟩_` is made available by importing `step-≡` because it is defined as an alternate syntax for it.

It's still not entirely clear to me why importing an operation/function should necessarily import its syntax operations as well, or why we can't just import the syntax we're interested in and have it import the function automatically. This is just because I don't know enough Agda, yet.

GetContented:
> It's still not entirely clear to me why importing an operation/function should necessarily import its syntax operations as well, or why we can't just import the syntax we're interested in and have it import the function automatically. This is just because I don't know enough Agda, yet.

Apologies for not making the implicit question here more explicit:

@wadler I'd like to be guided by you here on whether we should need to explain how Apda syntax importing works for the curious, or even to simply to refer to the Agda documentation by hyperlink. If we do that, your suggested briefer rewrite would be good. If you think a link would be good, I'd welcome what that link should be (given how green I am to the ecosystem).

Let me know your thoughts if you're able, and I'll implement whatever you think in my PR.

wenkokke:
I’m afraid that such a simple fix would still be confusing, but that a complete explanation would break up the flow of the text. I suggest that we acknowledge in the text that something strange is going on, and that `_≡⟨_⟩_` is imported by its alias `step-≡`, and that we link to a separate page which goes into the details of Agda’s syntax macros and why we have to have `step-≡` in the first place.

I don’t know whether we should write this explanation as a separate, unlisted “chapter”, or simply link to the documentation of syntax macros and the discussion—[on a GitHub issue](https://github.com/agda/agda-stdlib/issues/622) or in [the CHANGELOG](https://github.com/agda/agda-stdlib/blob/2d38f315b6a843f03c694c18c3e789593087ed87/CHANGELOG.md#changes-to-how-equational-reasoning-is-implemented)—of `step-≡`, but I would be leaning towards the former.

That is, I would recommend that we create a separate file…
```
src/plfa/notes/ReasoningSyntax.lagda.md
```
…in which we provide an overview of the problem, which we place in the appendix, and otherwise only link from this section and—later on—from the chapter on equality.

@wadler Thoughts?

GetContented:
Closing in preference of whatever others might suggest.

wenkokke:
@GetContented if you would be happy to work out my suggestion into a PR, I would be happy to merge it.

wadler:
I'm not sure we need a whole appendix to explain this point. Syntax declarations are described later in the book.

wenkokke:
Not so much a whole appendix, but a short aside explaining the step constructor, which happens to be linked in the appendices.

PR #652, opened 2022-05-22
WIP: upgrade to GHC-8.10.7, drop pandoc-citeproc

andreasabel:
Fixes #649 (CI).

CI currently looks very sad.  It basically never ran successfully.

This PR gets CI into a working state by:
- Upgrading from GHC 8.10.5 to 8.10.7 and the latest stack LTS 18.28 for GHC 8.10
- Dropping deprecated `pandoc-citeproc`, allowing newer `pandoc`
- Skipping the `processCites` step in page building until solution independent of `pandoc-citeproc` is available
- Pre-building the Agda standard library, fixing races in the site building process (locked `.agdai` files)
- Not fixing the CI branch to `dev`, allowing it to run on PRs like this one.
  (So, you may not see CI running on this PR before it has been merged into `dev`.)

In the current state (2022-05-21), CI succeeds (see https://github.com/andreasabel/plfa.github.io/actions/runs/2364301888), but there are the following loose ends / TODOs left:
- [ ] Clean up, comment on the decisions
- [ ] Extend CI to other platforms (Windows(?), add back macOS)
- [ ] Cache cabal directory
- [ ] Maybe cache Agda standard-library
- [ ] ~~Build Agda faster by turning off aggressive optimization~~ [EDIT: not easy; flag `optimise-heavily` was only introduced in Agda 2.6.2] (this would be obsolete by switching to Agda-2.6.2 which has it off by default)
- [ ] Migrate the processing of citations
- [x] Fix races in site building (ADDED after CI run)
  These might be fixed in the newer builders (shake/shoggoth)!?

Ad "migrating `processCites`":
- Atm, there are only two references for self-citations, thus, I decided skipping these would not be terrible: https://github.com/plfa/plfa.github.io/blob/a798dc6a76ced81fc3fbb869861d010423e30ccb/bib/plfa.bib
- I asked for migration help upstream: 
  * https://github.com/jgm/pandoc/issues/8081

Ad "races" (ADDED):
CI still experiences races (https://github.com/plfa/plfa.github.io/runs/6543439794?check_suite_focus=true#step:19:443) which I tried to circumvent by pre-building the Agda standard library:
```
/home/runner/work/plfa.github.io/plfa.github.io/src/plfa/part3/Soundness.lagda.md:50,1-63
/home/runner/work/plfa.github.io/plfa.github.io/_build/2.6.1.3/agda/src/plfa/part2/Substitution.agdai:
openBinaryFile: resource busy (file is locked)
```

[...]

PR #656, opened 2022-06-01
Fix module headers in part2.fixpoint-old.*

andreasabel:
See 
- https://github.com/plfa/plfa.github.io/pull/652#issuecomment-1143553143

Agda files saved to `fixpoint-old/` are missing `.fixpoint-old` as part of their module name.  Fixed here.

Please merge via "Merge commit" (not squash or rebase).


wenkokke:
@wadler What are these chapters? These don’t seem like they should be in the live version of the book. Shouldn’t they be moved to `extra`? What about the—presumably new—fixpoint files?

PR #723, opened 2022-07-07
change "embedding" to "retraction"

mikeshulman:
As discussed on the mailing list.


wenkokke:
Closing for now due to inactivity.

mikeshulman:
Was that the conclusion of the email discussion, to stick with "embedding"?

If there was too much pushback against "retraction", how about the alternative suggestion "split embedding"?

wenkokke:
I believe that embedding is fine as long as we define it locally within the book. However, I think it would be a good idea to mention the terms that might be used in other book, with references to those books

PR #724, opened 2022-07-07
Remove feed-validator from dev dependencies

fangyi-zhou:
Feed-validator is not used in the Makefile, not maintained (last release
5 years ago, repo archived), and there are unresolved vulnerabilities in
transitive dependencies.


wenkokke:
Yeah, and it broke recently. Good call.

PR #733, opened 2022-07-17
Bisimulation: environment -> context

fangyi-zhou:
In Bisimulation, there are two occurrences of "environment", I believe
the correct term is "context".
In the context of PL theory in general, they are interchangable.
However, I think for it is good to be pedantic in a textbook, and that
it's good to disambiguate that from "environments" in the BigStep
chapter.

Signed-off-by: Fangyi Zhou <me@fangyi.io>


wadler (commented):
Wen prefers the following terminology:
  context: a map from identifiers to types
  environment: a map from identifier to values, indexed by a context
By those definitions, environment is correct where used, and context is incorrect.

In fact, I am considering using "type environment" where the book currently uses "context", in order to be able to use context as in "evaluation context".

To match the terminology in Chapter De Bruijn, I think the two instances of "environment" should be "renaming" and "substitution", respectively.

Please just leave the terminology as it is for now, I will try to update it when I get around to adjusting terminology as mentioned above.

PR #764, opened 2022-09-28
fix code block delimiter in Soundness

keltono:
Fixes a minor issue where an inline code block is closed before intended, causing the some of the code to be outside of a code block and some of the prose to be inside of a code block.


jsiek (approved):
Thanks!

PR #789, opened 2022-11-18
TSPL: add my guest talk on Monday

omelkonian:


wadler:
Thanks, for both the talk and the pull request.

PR #799, opened 2022-12-08
Minor Fix: Remove Wrong `suc` in Explanation on Existential Quantifier Example

pmbittner:
Hello there,

In the chapter on quantifiers, in the subsection "_An existential example_", in the explanation of the third case of the proof of `odd-∃`, it is written:

    We return a pair consisting of `suc m` and evidence that `1 + m * 2 ≡ suc n`, which is immediate after substituting for `n`,

while actually `m` and not `suc m` is returned as the first element in the pair.


wadler:
Thank you!

PR #800, opened 2022-12-09
A unicode character table in appendix

bubblepipe:
I generated `src/plfa/part2/UnicodeCharTable.md` using the following shell script: 

```
#!/bin/bash

dirs=( "src/plfa/part1" "src/plfa/part2" )

for dir in ${dirs[@]} ; do

    for f in `ls $dir | grep .lagda.md`; do
    
        echo "#" $f
        echo
    
        sed -n -e '/## Unicode/,$p' $dir/$f
        echo
        echo
        echo
    done
done
```

And also updated `data/tableOfContents.yml` so this file shows in the appendix section. 

I did not build and test the changes locally though, as I do not have nodejs environment setup at this moment.

### DIFF
--- a/data/tableOfContents.yml
+++ b/data/tableOfContents.yml
@@ -43,6 +43,7 @@ part:
   title: "Appendix"
   chapter:
   - include: src/plfa/part2/Substitution.lagda.md
+  - include: src/plfa/part2/UnicodeCharTable.md
     epub-type: appendix
 - title: "Back matter"
   chapter:
--- a/src/plfa/part2/UnicodeCharTable.md
+++ b/src/plfa/part2/UnicodeCharTable.md
@@ -1,0 +1,346 @@
+# Connectives.lagda.md
+
+## Unicode
+
+This chapter uses the following unicode:
+
+    ×  U+00D7  MULTIPLICATION SIGN (\x)
+    ⊎  U+228E  MULTISET UNION (\u+)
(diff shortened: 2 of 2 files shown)


wenkokke:
I'm hesitant to merge this, because if we add a unicode table to the appendix, I'll have to maintain it, and therefore the bash used to generate it.

I'd be happy to merge this if the script were rewritten in Haskell, so I can maintain it as a part of the build system, and not as an additional bash script.

bubblepipe:
I am making this merge request primarily for the TSPL exam tomorrow, it would be great if you could approve it temporarily and undo it after tomorrow 3pm (UTC+0)?

I will attempt to rewrite the script in haskell after the exam.

wadler:
Thank you for putting together the Unicode table.

The suggestion to approve temporarily is reasonable, but unfortunately the submission failed some tests. (See above.) So I cannot accept it before Monday, sorry.

If you provide it with a Haskell script (and minus whatever error caused the test to fail) that would be great.

Copying from the chapters makes sense, but probably best to present them in the order they appear in the table of contents rather than in alphabetic order.

wenkokke:
Another issue here is that unfortunately we have no tests for the Unicode sections to trst that they’re complete, so picking them from these sections—while useful—isn’t necessarily correct or complete.

wenkokke:
I’ll close this for now, but, @AOIDUO, feel free to reopen!

PR #818, opened 2023-01-25
"index" instead of "parameter"

msvbg:
Is this a typo or am I wrong about the terminology?

### DIFF
--- a/src/plfa/part1/Lists.lagda.md
+++ b/src/plfa/part1/Lists.lagda.md
@@ -65,13 +65,13 @@ data List′ : Set → Set where
   []′  : ∀ {A : Set} → List′ A
   _∷′_ : ∀ {A : Set} → A → List′ A → List′ A
 ```
-Each constructor takes the parameter as an implicit argument.
+Each constructor takes the index as an implicit argument.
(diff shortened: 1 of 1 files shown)


wenkokke:
A "parameter" is any argument in a data definition that comes *before* the colon, e.g.,
```agda
-- Both `a` and `A` are parameters
data List {a : Level} (A : Set a) : Set a where
-- So we don't need a `forall {a : Level} (A : Set a)` when defining constructors
  nil : List A
  cons : A -> List A -> List A

-- The `a` is a parameter, but the `A` is an argument
data List {a : Level} : (A : Set a) -> Set a where
-- So we need a `forall (A : Set a)` when defining constructors
  nil : forall (A : Set a). List A
  cons : forall (A : Set a). A -> List A -> List A

-- Neither `a` nor `A` is a parameter
data List : {a : Level} (A : Set a) -> Set a where
-- So we need a `forall {a : Level} (A : Set a)` when defining constructors
  nil : forall {a : Level} (A : Set a). List A
  cons : forall {a : Level} (A : Set a). A -> List A -> List A
```

msvbg:
Thanks, I got confused!

PR #833, opened 2023-02-25
Update Agda version in README to 2.6.3

benbellick:
Current instructions may result in an 'agda-mode compile' build error. Updating the version resolves the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ The easiest way to install Agda is using Cabal. PLFA uses Agda version 2.6.2.2. 
 
 ```bash
 cabal update
-cabal install Agda-2.6.2.2
+cabal install Agda-2.6.3
 ```
 
 This step will take a long time and a lot of memory to complete.
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ or using `ghcup tui` and choosing to `set` the appropriate tools.
 
 ### Install Agda
 
-The easiest way to install Agda is using Cabal. PLFA uses Agda version 2.6.2.2. Run the following command:
+The easiest way to install Agda is using Cabal. PLFA uses Agda version 2.6.3. Run the following command:
 
 ```bash
 cabal update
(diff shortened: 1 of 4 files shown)


benbellick:
Should I also remove the "Build with newer Agda 2.6.3" job? 

https://github.com/plfa/plfa.github.io/blob/51c71b2ad10fa7b0d921bb11fd3a5553588f012c/.github/workflows/build.yml#L55

wenkokke:
> Should I also remove the "Build with newer Agda 2.6.3" job?
> 
> https://github.com/plfa/plfa.github.io/blob/51c71b2ad10fa7b0d921bb11fd3a5553588f012c/.github/workflows/build.yml#L55

It should be turned into a job building with the now older version 2.6.2.2.

These should also be updated:
https://github.com/plfa/plfa.github.io/blob/8d56d6fd23952243711449e6c80fa5ba713339c4/.github/workflows/build.yml#L60-L76

README.md
```diff
@@ -63,7 +63,7 @@ The easiest way to install Agda is using Cabal. PLFA uses Agda version 2.6.2.2.
 
 ```bash
 cabal update
-cabal install Agda-2.6.2.2
+cabal install Agda-2.6.3
```
wenkokke, line 66: The text above should also be updated to say 2.6.3.

wenkokke:
The following places also require updating.

The Agda version badge:
https://github.com/plfa/plfa.github.io/blob/dev/README.md?plain=1#L265-L266

And the links to the Agda documentation:
https://github.com/plfa/plfa.github.io/blob/dev/README.md?plain=1#L275-L280

Finally, the Agda standard library should be updated to 1.7.2 to match the Agda version.

This requires updating the badge:
https://github.com/plfa/plfa.github.io/blob/dev/README.md?plain=1#L267

As well as running the following command from the `standard-library` sub directory:
```
git checkout v1.7.2
```

wenkokke:
Everything looks good, except I don't think you committed the submodule update after running `git checkout v1.7.2`?

Edit: nevermind, you totally did, and the tests are failing because of the new standard live version.

wenkokke:
The standard library 1.7.2 seems to be backwards incompatible with Agda 2.6.2.2 and before, so we'll have to remove those tests. That's a little saddening, and might lead to some issues for users.

I've half a mind to remove the standard library from the repository and set it up with `setup-agda` so we can continue testing with older Agda versions.

wenkokke:
Pushed an attempt to adjust the stdlib version as needed, and fixed a few missing items.

wenkokke (approved):

wenkokke:
Thanks for helping out, @benbellick !

benbellick:
For sure thanks for guiding me! Hope I can continue to contribute moving forward 👍

PR #844, opened 2023-03-02
Reasoning and Example on Modelling Predicates as Inductive Data Types in Chapter `Equality` (#760)

### LINKED ISSUE #760: Possibly Sparse Description on `Set` and Predicates
Dear Agda-Team,

First of all, thanks for this awesome book and language! The book is a pleasure to read and doing the exercise within the book itself works like a charm and is a lot of fun. 😃

When facing Chapter 4 (Equality) of Part 1, I had trouble following the use and definition of predicates. In particular, the definition of substitution
```agda
subst : ∀ {A : Set} {x y : A} (P : A → Set)
  → x ≡ y
    ---------
  → P x → P y
subst P refl px = px
```
is the first occurence of predicates as parameters (of type `A → Set`). This is also the first time `Set` is used as a type. Previously, `Set` was used and introduced only in terms of `T : Set`, as a way to describe that `T` is a type in Agda (in Chapter 1 (Naturals) of Part 1). So it was unclear to me, how to interpret the use of `Set` within a function signature.

Moreover, the signature `P : A → Set` collided with my usual understanding of predicates as a function `A → Bool`. While the second signature `A → Bool` models predicates as functions, the first one `A → Set` models predicates as (dependent) types. This mental hop was a challenge to me as an apprentice of dependent types and is not explained or suggested within the chapter. (I am also unsure whether I fully understood it by now and got it right here.)

Potentially, it would have helped me as a reader of the book to

1. have an explicit explanation about the meaning of `Set`. I suppose it is the type of all types which is itself a type, giving rise to levels mentioned first in the paragraph on `Leibniz equality`? Is `Set` the type denoted by _U_ in [_The Little Typer_](https://thelittletyper.com/)? While there might be an explanation of `Set` in future chapters, this was the point where I was lost (at first).
2. have explanations or guidance towards the mental hop of expressing predicates as types instead of functions. I think `Even` and `Odd` are good examples here, which are predicates we already saw in an earlier chapter. Another option could be having an exercise for navigating the reader towards making this observation (in case they are not too familiar with dependent types yet).

This is just a suggestion and every reader might come with a different background of course.

Kind regards,
Paul

pmbittner:
Closes #760 .

Following suggestions in the discussion on issue #760, this pull request adds
- some reasoning on why predicates are modelled as parameterized types `A → Set`,
- a pointer to earlier examples `even` and `odd`,
- and a pointer to the future discussion on modelling predicates as types `A → Set` versus functions `A → Bool` in Chapter Decidable

at the first explicit occurence of generic predicates as paramaters in Chapter `Equality`.

I am happy to receive any feedback. 😃 (I had some trouble setting up ruby to test the new links, so they are yet untested; but I formatted them according to existing links in the book.)

Kind regards,
Paul

### DIFF
--- a/src/plfa/part1/Relations.lagda.md
+++ b/src/plfa/part1/Relations.lagda.md
@@ -649,7 +649,7 @@ the fact that inequality is transitive.
 ```
 
 
-## Even and odd
+## Even and odd {#even-and-odd}
 
 As a further example, let's specify even and odd numbers.  Inequality
 and strict inequality are _binary relations_, while even and odd are
--- a/src/plfa/part1/Relations.lagda.md
+++ b/src/plfa/part1/Relations.lagda.md
@@ -649,7 +649,7 @@ the fact that inequality is transitive.
 ```
 
 
-## Even and odd {#even-and-odd}
+## Even and odd
 
 As a further example, let's specify even and odd numbers.  Inequality
 and strict inequality are _binary relations_, while even and odd are
--- a/src/plfa/part1/Equality.lagda.md
+++ b/src/plfa/part1/Equality.lagda.md
@@ -171,7 +171,12 @@ subst : ∀ {A : Set} {x y : A} (P : A → Set)
   → P x → P y
 subst P refl px = px
 ```
-
+A predicate is a proposition over values of some type `A`, and since we model
+_propositions as types_, a predicate is a type parameterized in `A`.
+As an example, consider our earlier examples `even` and `odd` from
+Chapter [Relations](/Relations/#even-and-odd), which are predicates on natural numbers `ℕ`.
+(We will compare representing predicates as inductive data types `A → Set`
+versus functions to booleans `A → Bool` in Chapter [Decidable](/Decidable/).)
 
 ## Chains of equations
 


src/plfa/part1/Relations.lagda.md
```diff
@@ -649,7 +649,7 @@ the fact that inequality is transitive.
 ```
 
 
-## Even and odd
+## Even and odd {#even-and-odd}
```
wenkokke: This shouldn't be necessary. That anchor is automatically generated.

wenkokke:
This looks good to me.

@wadler?

src/plfa/part1/Relations.lagda.md
pmbittner: Ok. I guess the generated name will be `#Even-and-odd`?

wenkokke: I think it removes all capitalization, but it should be easy to check on the live webpage.

pmbittner: True! I just checked, it is `#even-and-odd`. 😃

wadler (commented):
Thanks!

PR #871, opened 2023-05-29
Update Lists.lagda.md

basetunnel:
Change non-breaking space to normal space. It was rendered as a kappa in the book.


wadler:
Good catch! Thanks, Jacco.

PR #876, opened 2023-05-30
(typo) if this "and" were "on" it would confuse me less

williamdemeo:


wenkokke:
The pattern and the right hand side of the equation are different things, so this would be wrong.

wadler:
Thanks!

williamdemeo:
I completely agree that "the pattern" and "the right hand side" are different things. 

However, "the pattern" is to be matched with "the expression" *on* "the right hand side"--NOT the expression *and* the right hand side.

I think you are confused and should not have closed this PR.

PR #878, opened 2023-05-31
Rename the B in existential syntax

ch1bo:
We discussed this in our course today and using B_x makes it clearer and not conflict with the data declaration of existentials ("These are two different Bs").

### DIFF
--- a/src/plfa/part1/Quantifiers.lagda.md
+++ b/src/plfa/part1/Quantifiers.lagda.md
@@ -143,7 +143,7 @@ We define a convenient syntax for existentials as follows:
 ```agda
 Σ-syntax = Σ
 infix 2 Σ-syntax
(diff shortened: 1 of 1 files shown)


wenkokke:
This would likely require a new unicode entry at the end of the chapter.

Is there a simpler name, like BC or BxC, to show that it is a pair?

wadler:
Using B sub x is correct. Thanks, Sebastian! Please add Unicode to the appendix, as suggested by Wen.

wenkokke:
Ah, it's "B containing x".  Should've clocked that. We could opt for `Bx` or `B[x]` and avoid more unicode?

ch1bo:
Renamed it to `Bx` now, but I'm fine with anything that is not `B` to make it a bit less ambiguous for the reader (who might not know the distinction between the B used in the definition and the usage as variable in the syntax)

PR #881, opened 2023-06-02
Add a tips page to the book

michaelpj:
This is a collection of tips I picked up during the workshop.


wadler:
Thanks, Michael!

(Wen, I asked Michael to make this pull request.  He has a useful tips file, and it makes sense to add it to the book as backmatter. I hope others will submit new tips.)

michaelpj:
Hmm, something is unhappy but I'm not sure what.

src/plfa/backmatter/Tips.md
```diff
@@ -0,0 +1,197 @@
...
+This will solve the expression holes at the beginning and the end so you don't have to fill them in.
+You can do this again if you fix a step, so if you fill in the reasoning you want to do for a step, you can use "solve" and Agda will fill in the next term.
+
+## Goal information will show computed terms
+
+Suppose you have a partial equational reasoning proof:
+```agda
+begin
+    2 + (x + y)
+  ≡⟨ ?2 ⟩
```
wenkokke, line 36: This probably shouldn't have a `2`? Or, if it does, why `2`?

src/plfa/backmatter/Tips.md
```diff
@@ -0,0 +1,197 @@
...
+    ?
+∎
+```
+
+The first term can be computed more.
+To see this, you can ask for the goal context for `?2` and it will show something like this:
+```
+?2 = suc (suc (x + y)) = y_5244
+```
+i.e. your LHS and an unknown RHS (because it's a hole), but it will show the LHS fully computed, which you can copy into your file directly.
```
wenkokke, line 50: You may need to disambiguate LHS and RHS for the reader.

src/plfa/backmatter/Tips.md
```diff
@@ -0,0 +1,197 @@
...
+```agda
+foo a with my-thing
+... | x = ?
+```
+and then ask to case-split on `x`.
+
+## Quickly setting up record skeletons
+
+"Refine" (`C-c C-r` in Emacs) will set up a skeleton for a record constructor if Agda knows a hole must be a record type.
+This is useful for e.g. isomorphisms, products, sigmas, or existentials.
```
wenkokke, line 69: Maybe also add defining records via their projections, and the corresponding `C-c C-c ENTER`?

src/plfa/backmatter/Tips.md
```diff
@@ -0,0 +1,197 @@
...
+```agda
+foo b * 2
+```
+
+The definition of `*` does case analysis on its first argument.
+This means that if the first argument to `*` is a constructor application, Agda can do some computation for you.
+The first example computes to `foo b + (foo b + zero)`, but the second one does not compute at all: Agda can't pattern match on `foo b`, it could be anything!
+
+Doing this lets Agda do more definitional simplification, which means you have to do less work.
+Unfortunately it does require you to know _which_ arguments are the one that Agda can evaluate, which requires looking at the definitions of the functions.
```
wenkokke, line 89: There's some general advice we can probably give here to help people guess which arguments are evaluated, e.g., left associative operators tend to evaluate their left argument.

src/plfa/backmatter/Tips.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+They use three approaches for doing "multiple steps":
+1. Equational reasoning
+2. `trans`
+3. `rewrite`
+
+However, notably they all use the same supporting proofs!
+That means you can often write the overall proof however you find easiest and then rewrite it into another form if you want.
+For example, do the proof slowly with equational reasoning, and then turn it into a compact proof with `rewrite`.
```
wenkokke, line 136: I'd hesitate to advise folks to rewrite their readable equational reasoning proofs into less readable rewrite ones.

I'd ask that you remove that particular tip.

(It's also not obvious that such a rewrite always works.)

src/plfa/backmatter/Tips.md
```diff
@@ -0,0 +1,197 @@
...
+There are a few ways to avoid green slime, but one way that often works is to use a variable and also include a proof that it has the value that you want. So:
+
+```agda
+data Tree (A : Set) : ℕ → Set where
+  leaf : A → Tree A 1
+  node :
+    ∀ {s : ℕ} {s‵ : ℕ}
+    → Tree A s
+    → Tree A s
+    → s‵ ≡ s + s
```
wenkokke, line 193: This should be a regular quote or a prime, but this seems to be a backtick?

wenkokke:
Fixed the issue. You were missing an `epub-type` and, apparently, triggering some obscure Pandoc 2.19 bug?

src/plfa/backmatter/Tips.md
```diff
@@ -0,0 +1,197 @@
...
+    ?
+  ≡⟨ ? ⟩
+    ?
+∎
+```
+
+This sets up a series of reasoning steps with holes for both the expressions at each step, and the step itself.
+Since the reasoning steps are holes, Agda will always accept this, so you can make incremental progress while keeping your file compiling.
+
+Now call "solve" (`C-c C-s` in Emacs).
```
buggymcbugfix, line 26: I run `C-c C-a` (`auto`) instead, which has the added benefit of getting rid of the hole in a single step. (At least in vscode I need to `C-c C-space` to `give` after solving.)

src/plfa/backmatter/Tips.md
buggymcbugfix: You can also just remove `?2` and surrounding spaces (or replace it with `refl`) and then auto on the next `?` below.

src/plfa/backmatter/Tips.md
buggymcbugfix: Perhaps tell people about jump-to-definition?

src/plfa/backmatter/Tips.md
```diff
@@ -0,0 +1,197 @@
...
+They use three approaches for doing "multiple steps":
+1. Equational reasoning
+2. `trans`
+3. `rewrite`
+
+However, notably they all use the same supporting proofs!
+That means you can often write the overall proof however you find easiest and then rewrite it into another form if you want.
+For example, do the proof slowly with equational reasoning, and then turn it into a compact proof with `rewrite`.
+
+## Avoid mutual recursion in proofs by recursing outside the lemma
```
buggymcbugfix, line 138: I think I understand what this advice is trying to say, but it's a bit unclear. The title just adds to the confusion.

I think what you are trying to say here is that we should try to break up a proof into small lemmas, that do one reasoning step, which we then compose to get the overall goal?

src/plfa/backmatter/Tips.md
```diff
@@ -0,0 +1,197 @@
...
+≤-refl : ∀ { n : ℕ } → n ≤ n
+```
+Agda will be able to infer the `n` from the `n ≤ n` in the result.
+
+Note that the situation is different for constructors and functions.
+Generally, Agda can infer `x` from `f x` if `f` is a constructor (like `≤`) but _not_ if it is a function (like `_+_`).
+
+## Avoiding "green slime"
+
+"Green slime" is when you use a function in the index of a returned data type value.
```
buggymcbugfix, line 174: Maybe some explanation why "green slime"?

buggymcbugfix: Further reading here: https://proofassistants.stackexchange.com/a/1615 (There is also a link to Connor's paper that coined this term.)

noughtmare:
I think this is such an important resource, so I'm very sad to see this has stalled. What can I do to help push it over the finish line?

noughtmare:
I've addressed the comments here in a new PR: #1137

PR #906, opened 2023-09-05
remove misleading sentence about the origin of the proof

jsiek:


wenkokke:
It seems a lot of this passage could benefit from using citations. Do you have the bibTeX entries for these, @jsiek?

jsiek:
I'm sure I can find some. I'll look into that.

PR #933, opened 2023-11-13
Bump GHC and Cabal versions

wenkokke:


OlingCat:
I have the same error:
```
Failed to build charsetdetect-1.1.0.2. The failure occurred during the
configure step.
Build log (
D:\PL\ghcup\cabal\logs\ghc-9.4.7\charsetdetect-1.1.0.2-cb2f0cc42c64b6ebf62b5b873d0f0c8068f81202.log
):
Warning: charsetdetect.cabal:66:51: version operators used. To use version
operators the package needs to specify at least 'cabal-version: >= 1.8'.
Configuring charsetdetect-1.1.0.2...
cabal-3.6.2.0-p1.exe: Missing dependencies on foreign libraries:
* Missing (or bad) C libraries: stdc++-6, gcc_s_seh-1
This problem can usually be solved by installing the system packages that
provide these libraries (you may need the "-dev" versions). If the libraries
are already installed but in a non-standard location then you can use the
flags --extra-include-dirs= and --extra-lib-dirs= to specify where they are.If
the library files do exist, it may contain errors that are caught by the C
compiler at the preprocessing stage. In this case you can re-run configure
with the verbosity flag -v3 to see the error messages.

cabal-3.6.2.0-p1.exe: Failed to build charsetdetect-1.1.0.2 (which is required
by exe:builder from plfa-22.8). See the build log above for details.

make: *** [Makefile:34: build] Error 1
```

So I copied `libgcc_s_seh-1.dll` and `libstdc++-6.dll` from my own mingw installation into `D:\PL\ghcup\ghcup\ghc\9.4.7\mingw\bin` but didn't work.  Do you have any idea?

wenkokke:
I've noticed charsetdetect breaking on Windows frequently, and the only solution I've found to work consistently is not to use it.

OlingCat:
> I've noticed charsetdetect breaking on Windows frequently, and the only solution I've found to work consistently is not to use it.

Will you replace charsetdetect in subsequent version?

wenkokke:
It's a bit of a tragic bargain. It breaks on Windows, but assuming files are UTF-8 encoded had a tendency to... break things on Windows.

wenkokke:
I guess the answer is yeah, though.

wenkokke:
I guess the answer is I try to patch charsetdetect. And it sort of works, but now there's an issue with the cabal project configuration.

wenkokke:
@OlingCat now fixed, I hope.

PR #957, opened 2024-01-31
Add reference to Unicode section

buggymcbugfix:
A little in-place reminder for those who already garbage collected the contents of the "Getting Started" chapter.

EDIT: I put it as part of the first exercise as that should be the first time the issue of inputting special characters arises for anyone working through the book.

### DIFF
--- a/src/plfa/part1/Naturals.lagda.md
+++ b/src/plfa/part1/Naturals.lagda.md
@@ -83,6 +83,8 @@ Write out `7` in longhand.
 
 You will need to give both a type signature and definition for the
 variable `seven`. Type `C-c C-l` in Emacs to instruct Agda to re-load.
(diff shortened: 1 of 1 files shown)


buggymcbugfix:
My bad, I just noticed that several paragraphs further down, there is the following explanation:


> You may have noticed that `ℕ` and `→` don't appear on your keyboard.
They are symbols in _unicode_.  At the end of each chapter is a list
of all unicode symbols introduced in the chapter, including
instructions on how to type them in the Emacs text editor.  Here
_type_ refers to typing with fingers as opposed to data types!

This is a good and more thorough explanation that what I have in this PR, but I daresay it is not well-placed. Could we move this explanation up, near the first exercise?

wenkokke:
Moved to issue #958

Issue #965, opened 2024-02-21
Type mismatch when `make build`


OlingCat:
I tried to `make build` PLFA and encountered the following error:

```
$ make build
Building PLFA...
cabal  v2-update
Downloading the latest package list from mirrors.ustc.edu.cn
Package list of mirrors.ustc.edu.cn is up to date.
The index-state is set to 2024-02-20T21:39:04Z.
cabal  v2-run builder -- build -j --lint --profile=_cache/reports/build.html --timing
Build profile: -w ghc-9.4.8 -O1
In order, the following will be built (use -v for more details):
 - plfa-22.8 (exe:builder) (first run)
Preprocessing executable 'builder' for plfa-22.8..
Building executable 'builder' for plfa-22.8..
[1 of 1] Compiling Main             ( tools/Buildfile.hs, /mnt/d/Source/plfa/dist-newstyle/build/x86_64-linux/ghc-9.4.8/plfa-22.8/x/builder/build/builder/builder-tmp/Main.o )

tools/Buildfile.hs:273:28: error:
    • Couldn't match type ‘base64-1.0:Data.Base64.Types.Internal.Base64
                             'base64-1.0:Data.Base64.Types.Internal.StdPadded LazyText.Text’
                     with ‘LazyText.Text’
        arising from a functional dependency between constraints:
          ‘?getDigest::FilePath -> Action (Maybe LazyText.Text)’
            arising from a use of ‘getStylesheetField’ at tools/Buildfile.hs:273:28-45
          ‘?getDigest::FilePath
                       -> Action
                            (Maybe
                               (base64-1.0:Data.Base64.Types.Internal.Base64
                                  'base64-1.0:Data.Base64.Types.Internal.StdPadded LazyText.Text))’
            arising from the implicit-parameter binding for ?getDigest at tools/Buildfile.hs:229:7-15
    • In a stmt of a 'do' block: stylesheetField <- getStylesheetField
      In the expression:
        do src <- Route.source out
           tocField <- getTableOfContentsField ()
           (fileMetadata, indexMarkdownTemplate) <- getFileWithMetadata src
           stylesheetField <- getStylesheetField
           ....
      In the second argument of ‘(%>)’, namely
        ‘\ out
           -> do src <- Route.source out
                 tocField <- getTableOfContentsField ()
                 ....’
    |
273 |         stylesheetField <- getStylesheetField
    |                            ^^^^^^^^^^^^^^^^^^
[...]

PR #968, opened 2024-02-26
Add upper bound for base64

wenkokke:


wenkokke:
Fix #965.

wenkokke:
Doesn't work with GHC 9.6.

PR #986, opened 2024-04-29
Suggested modification to first exercise

### LINKED ISSUE #960: First exercise too advanced?
The very first exercise in the book asks the reader to write code before any of the syntax has been presented:

> #### Exercise `seven` (practice) {#seven}
> 
> Write out `7` in longhand.
> 
> ```agda
> -- Your code goes here
> ```
> 
> You will need to give both a type signature and definition for the
> variable `seven`. Type `C-c C-l` in Emacs to instruct Agda to re-load.

Intuitively, reading the question, my answer would have been (suc (suc (suc (suc (suc (suc (suc zero))))))), but the next sentences then ask for a type signature and explain how to reload the code in Emacs. This makes me think that I do need to write code after all. 
Since the reader does not yet know how to write Agda definitions this may cause some frustration (it certainly did for me). 

An educational way of fixing this might be this:

> ```agda
> -- seven : ℕ
> -- seven = <Your code goes here>
> ```
>
> Uncomment the code by removing the dashes and insert your definition.

KevinDCarlson:
Fixes #960 by handing the student the type declaration. Starts with an explicitly wrong answer to avoid `?`s.


src/plfa/part1/Naturals.lagda.md
```diff
@@ -14,7 +14,7 @@ contains about seventy sextillion stars.
 But the number of stars is finite, while natural numbers are infinite.
 Count all the stars, and you will still have as many natural numbers
 left over as you started with.
-
+{-testing ∀ -}
```
wenkokke: Did you intend for this to be here?

src/plfa/part1/Naturals.lagda.md
```diff
@@ -75,14 +75,14 @@ successor of two; and so on.
 
 #### Exercise `seven` (practice) {#seven}
 
-Write out `7` in longhand.
+Write out `7` in longhand. The suggestion below loads but is, of course, incorrect.
```
wenkokke, line 77: Why do you want to avoid `?`?

KevinDCarlson: Because at this point, the reader hasn't learned how to deal with `?`--the walkthrough of that is a bit later.

src/plfa/part1/Naturals.lagda.md
KevinDCarlson: Whoops, nope. Just removed it.

src/plfa/part1/Naturals.lagda.md
wadler: Actually, holes are never used in the executable text, because Agda gets stroppy if you try to import a module that contains a hole.

wadler (approved):
Thanks!

PR #987, opened 2024-04-29
add option to `git clone` to prevent issue with recursive submodules

nrnrnr:
A recent vulnerability fix prevents `git clone` from working with git version 2.39.2.  This patch opens the vulnerability for the clone command only, meaning that the user is trusting the plfa repo.

Details at https://stackoverflow.com/questions/74486167/git-clone-recurse-submodules-throws-error-on-macos-transmission-type-file-n

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ If you'd like, you can [test to see if you've installed Agda correctly][agda-rea
 We recommend installing PLFA from Github into your home directory, by running the following command:
 
 ```bash
-git clone --depth 1 --recurse-submodules --shallow-submodules https://github.com/plfa/plfa.github.io plfa
+git -c protocol.file.allow=always clone --depth 1 --recurse-submodules --shallow-submodules https://github.com/plfa/plfa.github.io plfa
 ```
 
 PLFA ships with the required version of the Agda standard library, so if you cloned with the `--recurse-submodules` flag, you’ve already got it, in the `standard-library` directory!


wadler (commented):
Thanks, Norman!

A quick web search tells me the current version of git is 2.45.0.

Is it a good idea to weaken security for everyone to ensure compatibility with an old version of git?

I wonder if a better idea is to add a "troubleshooting" page where we can list problems and fixes. (I just had one similar submission asking to add a couple of sentences to help people with a problem that occurs on some Macs, but which I've never seen on my Mac. A troubleshooting page would be the right place for that to go too.)

What do you think?

nrnrnr:
(Quick reply as I'm getting ready for a thesis defense.)

The problem was introduced in git version 2.38 as a way of addressing CVE-2022-39253.  I don't know for what versions of git the workaround is necessary.  Not sure how to discover that---do you want to try with whatever version you are running?

I do know that Debian stable is affected, and the next thing I would check would be Ubuntu LTS.  2020 LTS is affected for sure, and likely also 2022 LTS.  Those are supposed to be supported for 5 years each.

The security vulnerability here is not quite clear to me, but the attack vector would be a fake DNS record for `github.com`.

nrnrnr:
I did a bit more research. As near as I can tell from [the description of the vulnerability](https://www.suse.com/security/cve/CVE-2022-39253.html), this issue is going to affect all git version from 2.38 onward.  So I'd say either update the instructions or find a way to use submodules that doesn't trigger the issue.  (I don't know from submodules, so I can't help with the latter.)

Also, the current git version for Red Hat Enterprise Linux is 2.39.3.

wenkokke:
Do we know what in the PLFA repository currently violates the new policy?

nrnrnr:
> Do we know what in the PLFA repository currently violates the new policy?

I don't understand the issue except that it somehow relates to submodules.

wenkokke:
@nrnrnr Does the workflow in #994 fix your problem?

nrnrnr:
> @nrnrnr Does the workflow in #994 fix your problem?

It clones without error, and I can compile the `Naturals` module, so I think we can declare that fixed!  Thanks.

PR #988, opened 2024-04-29
Add warning about `C-space` on Mac

KevinDCarlson:
Modern OS X seems to reserve `C-space` for toggling keyboard encodings. The VS Code plugin suggests `C-c C-r` as usually a reasonable replacement for `C-c C-space`, so I've suggested adding the same tip. I don't know enough about `C-c C-r` to know whether this is dangerous advice or needs more expansion, or whether it's worth trying to understand how to override the system's control of `C-space` entirely.

### DIFF
--- a/src/plfa/part1/Naturals.lagda.md
+++ b/src/plfa/part1/Naturals.lagda.md
@@ -838,6 +838,9 @@ filled, you can type `C-c C-space`, which will remove the hole:
     zero + n = n
     suc m + n = { }1
 
+Note: `C-space` may be a reserved command on Mac OS X, in which case
+try typing `C-c C-r` instead.
+
 Again, going into hole 1 and typing `C-c C-,` will display information on the
 required type of the hole, and what free variables are available:
 


wadler (commented):
Thanks for the suggestion!

I use a Mac, and I've never encountered that particular problem.

Rather than enlarge the text with a sentence or two for each problem some user might encounter, I wonder if it is better to add a "troubleshooting" section with hints such as this one. What do you think?

KevinDCarlson:
Yes, that seems fine. I’m confused about who exactly this affects as one reading of this post is that it ought to affect everyone. Possibly that was briefly true and now it only affects people with multiple keyboards enabled. In fact given that it doesn’t affect all Macs maybe this is fine to just let people Google. https://www.reddit.com/r/MacOSBeta/s/PhcqVbBc2f

JacquesCarette:
I think emacs overrides this - I've never had issues with `C-space` on Aquamacs. So it might be only `VSCode` that has this issue.

wenkokke:
I've not had this issue with VSCode.

wenkokke:
Closing for now. Feel free to reopen if you encounter this issue.

PR #990, opened 2024-05-03
Prevent color change of visited definitions in code blocks

### LINKED ISSUE #989: `a:visited` in code blocks messes up highlighting
The css has a general `a:visited` defined to a tint of blue, which applies to code blocks and thus changes highlighting:

code_block

Here `to`, `from`, `Set` and `extensionality` have been visited. This is less critical for function and type names but, as can be seen, changes the color for record fields (and would do for the `refl` type constructor as well) significantly.

adql:
Add a `a:visited` rule to `web/sass/theme/agda{-dark}.scss` files to override the general rule inside `.Agda` elements. Closes #989.


wenkokke:
You're setting the text color to transparent? 
Could you explain how that works and why it solves the issue?
My guess would be to just set the color to inherit for visited.

adql:
Apparently `transparent` [only works in Firefox](https://developer.mozilla.org/en-US/docs/Web/CSS/:visited#privacy_restrictions) (*sigh*) and it's indeed too much of a hack. But anyway the real issue is with specificity and there's no need to add an overriding `a:visited` rule.

wenkokke:
Does the latest commit address the issue? I'd expect to see at least a modification to the base Sass to exclude `.Agda`.

adql:
The latest commit addresses the issue on both Firefox and Chromium. `.Agda .{classname}` has specificity weight 0-2-0, while `a:visited` has 0-1-1. I can have a look on using an exclusion rule instead, but as I pointed out [in the corresponding issue](https://github.com/plfa/plfa.github.io/issues/989#issuecomment-2093237825) this is how the original Agda html backend styles the code blocks.

PR #994, opened 2024-05-04
Fix #987

wenkokke:


wadler (commented):
What is the reason for this change?
Plus: it avoids the scary `--recurse-submodules` and `--shallow submodules` commands.
Minus: I now need to cut-and-paste two commands instead of one.

wenkokke:
That's the proposed fix for the CVE noted in #987.

wenkokke:
The issue is with `--recurse-submodules` and the suggestion is to avoid it.

wadler:
Thanks. I followed the #987 link after I wrote my message, after which I wrote a second message beginning "Doh". Somehow the second message wasn't posted. Yes, I agree with the change.

PR #1002, opened 2024-05-27
Add catch-all explanation for users who installed Agda on Windows

DavidPratten:
The purpose of this commit is to ease the pathway for people installing Adga on Windows. Installing Agda on Windows using ghcup and cabal doesn't create an environment variable called AGDA_DIR and doesn't place this directory in the suggested location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,11 @@ cp ~/plfa/data/dotagda/* ~/.agda
 
 This provides access to both the Agda standard library and to PLFA as an Agda library.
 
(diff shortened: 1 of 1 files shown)


DavidPratten:
Further experimentation has shown that the folder returned by:
 
```bash
agda --print-agda-dir
```
is not the same place as `AGDA_DIR`.  `libraries` and `defaults` files created there are not found by agda.

Instead, on Windows, the directions already there are correct and creating `C:\Users\USERNAME\AppData\Roaming\agda` and placing  `libraries` and `defaults` gives the expected outcome.

I'm curious about the purpose of the `--print-agda-dir` command line option and how it relates to the `AGDA_DIR`?

wadler (commented):
Thank you!

wenkokke:
They're different directories. You can set the other directory with the `Agdadata_dir` variable. It's where Agda keeps some of its auxiliary files.

Issue #1013, opened 2024-07-12
Bug in appendix Substitution


wadler:
Section [The σ algebra equations](https://plfa.inf.ed.ac.uk/Substitution/#the-%CF%83-algebra-equations) contains the equation

    (sub-abs)   ⟪ σ ⟫ (ƛ N)    ≡ ƛ ⟪ σ ⟫ N

This is wrong, as is the subsequent discussion ("The equations sub-app and sub-abs says that substitution is a congruence for the lambda calculus. ") Later, the correct equation is stated and proved.

    sub-abs : ∀{Γ Δ} {σ : Subst Γ Δ} {N : Γ , ★ ⊢ ★}
            → ⟪ σ ⟫ (ƛ N) ≡ ƛ ⟪ (` Z) • (σ ⨟ ↑) ⟫ N

jsiek:
Thanks for noticing this! I've made a pull request to fix it.

noughtmare:
It seems this issue can be closed.

PR #1014, opened 2024-07-14
fix sub-abs in summary of the sigma algebra

jsiek:


jsiek:
fixes issue #1013

wadler:
Wow, that was fast!

PR #1020, opened 2024-07-26
motivate and explain parallel reduction, plus wordsmithing

nrnrnr:
This PR holds a small addition to the Confluence chapter.
It explains a bit more what can go wrong with the diamond
property and how that scenario motivates the definition of
parallel reduction.


wadler:
Jeremy is the primary author of Confluence, so he should review.

JacquesCarette:
@jsiek self-assigned this a year ago, and... forgot?

PR #1021, opened 2024-07-26
explain neutral terms and add a simple exercise

nrnrnr:
In the Untyped chapter, I did not
understand the motivation for neutral terms.  The best I
could find through web search was to notice that a neutral term
can be substituted for a variable without creating a β-redex.  But
the Untyped chapter does not mention this property.

To address my own concern, I have written something about
alternative ways to define normal forms.  And I have conjectured
that you prefer using neutral terms because it requires fewer Agda
constructors to specify normal forms─thereby reducing the number
of cases in relevant proofs.  The PR discusses
alternatives and also includes a couple of exercises.

The pull request includes this unanswered question: Why does
the Untyped chapter use syntactic criteria to define normal form,
when the earlier chapters on typed lambda calculi used a semantic
criterion (the term does not reduce)?

### DIFF
--- a/src/plfa/part2/Untyped.lagda.md
+++ b/src/plfa/part2/Untyped.lagda.md
@@ -292,7 +292,7 @@ of values, we are now interested in _normal forms_.
 [Why in this chapter do we introduce a syntactic criterion for
 normal form, whereas in the Properties chapter we define a normal form
 using a semantic criterion (as a term that does not reduce)?]
-A term is in normal form if it does not contain a β-redex: 
+A term is in normal form if it does not contain a β-redex:
 
   - A variable is in normal form.
 
@@ -370,7 +370,7 @@ neutral terms, and using `#′` in place of `#`
 #### Exercise (`non-abstraction`) (practice)
 
 Define a judgment `¬ƛ` mapping well-scoped terms to `Set`.
-Type `¬ƛ M` is inhabited if and only if `M` is not an abstraction. 
+Type `¬ƛ M` is inhabited if and only if `M` is not an abstraction.
 
 ```agda
 -- Your code goes here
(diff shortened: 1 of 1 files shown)


wadler (changes requested):
We cannot accept a pull request with a bracketed comment that is to be removed later. Please resolve such comments before putting in pull requests.

The revised text is much longer than the current text, and in my view less helpful.

I take it seriously when a reader doesn't understand a passage. That means something should change! But I'm not keen on a change that is much longer, and replaces my intuition by your intuition---best to have both, but I definitely don't want to lose my point of view.

I don't understand why terms that are not lambdas play a special role in normal forms or how that helps. A better exercise may be to show that terms in normal and neutral form do not reduce.

nrnrnr:
I didn't expect the PR to be accepted as is.   But I thought a PR the best way to start the discussion.

I don't mind removing the bracketed comment or writing the exercise (and a model solution).  But I can't express your intuition.  So until I hear from you that updating the PR would be productive, I plan to do nothing.

The unanswered questions are:

  - Why does specifying a normal form require two judgments, when in past chapters one judgment was sufficient to express Value?
  - Why is "neutral" a good second judgment?
  - How does "neutral" connect to the broader world of the study of the lambda calculus?

mskiptr:
@nrnrnr I'm just a passerby here, but I think you might find this information useful: When creating a GitHub pull request, it's possible to mark it as a draft. Then, the maintainers will hopefully have the right expectations from the very start.

nrnrnr:
> @nrnrnr I'm just a passerby here, but I think you might find this
 > information useful: When creating a GitHub pull request, it's possible to
 > mark it as a draft. Then, the maintainers will hopefully have the right
 > expectations from the very start.

Good to know; thanks.

PR #1025, opened 2024-08-28
document .gitignore. Add .DS_Store to it while at it.

JacquesCarette:
As per title.


JacquesCarette:
So it seems to pre-commit hook didn't like my changes to `.gitignore`. Are the rules to obey written down somewhere?

wenkokke:
The pre-commit setup sorts it alphabetically, which is incorrect given that .gitignore exclusions and inclusions are order-dependent, so that should probably be taken out.

(The details are in the pre-commit config file. Follow the links to the various repositories to find out what code each action runs.)

JacquesCarette:
This one is ready for review @wenkokke .

.gitignore
```diff
@@ -1,8 +1,18 @@
+# PLFA specific items to ignore
 !cabal.project.ci.local
 !data/legacy/19.08.zip
 !data/legacy/20.07.zip
 !plfa.pdf
```
wenkokke: These exclusions should be below their respective inclusions.

wenkokke (dismissed):

.gitignore
JacquesCarette: Done.

wenkokke:
I meant after their respective inclusions, eg, the zip inclusion after the general exclusion, but that's alright, we can fix that later.

PR #1057, opened 2024-11-03
Remove requirement for pattern matching in `×-comm` and `×-assoc`

hafniz:
Following 933ef88a259e6e7c0bb7dd54122ca57ef5d4c564, functions `from∘to` and `to∘from` in `×-comm` and `×-assoc` do not pattern-match the `⟨_,_⟩` structure of conjunction type anymore, but the accompanying text explanation has not been updated. I removed the requirement for pattern matching in the text to be consistent with the code. 

Also fixed a typo that calls the constructor for `⊤` `tt′` - should be `tt`.

-Haofei


wadler (approved):
Thank you!

PR #1061, opened 2024-11-05
Fix in imports for assignment3 2024

vsrivatsa-edinburgh:
Fixed import for ¬?


wadler (commented):
Thank you. I already made similar changes, and they are already published on GitHub and at plfa.inf.ed.ac.uk.

PR #1066, opened 2024-11-13
Use double substitution for alternative formulation of products

hafniz:
Also fixing the format issue under 'typing' for sums

-Haofei


wadler (approved):
Thank you!

PR #1071, opened 2024-11-21
Update Lambda.lagda.md Diamond and confluence example

b-m-f:
The diamond property written in Agda demand a single step from L->(M|N)?  Therefore the verbose text should reflect the picture, which has 3 lines at the top.

### DIFF
--- a/src/plfa/part2/Lambda.lagda.md
+++ b/src/plfa/part2/Lambda.lagda.md
@@ -791,10 +791,10 @@ It can be illustrated as follows:
                P
 
 Here `L`, `M`, `N` are universally quantified while `P`
-is existentially quantified.  If each line stands for zero
+is existentially quantified.  If each of the four lines in the figure above stands for zero
 or more reduction steps, this is called confluence,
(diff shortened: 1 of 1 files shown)


wadler (changes requested):
Your fix is incorrect. Please submit the following fix instead. Replace

"If each line stands for zero or more reduction steps, this is called confluence, while if the top two lines stand for a single reduction step and the bottom two stand for zero or more reduction steps it is called the diamond property." 
-->
"If each line in the figure above stands for zero or more reduction steps, this is called confluence, while if the top two lines stand for a single reduction step and the bottom two stand for zero or more reduction steps it is called the diamond property."

b-m-f:
OK,
I now see that the lines are meant as the 1 line from L to M and 1 line from L to N.

I saw each direction as a combination of 3 lines, as the individual characters.

b-m-f:
I have updated the change. What do you think about making it clear that the individual characters `\`, `/` connect to a single line and the figure has 4 lines total?

PR #1080, opened 2025-01-08
Add imports for *-mono-<= exercise

### LINKED ISSUE #1079: Import _*_ in Relations
I just did exercise *-mono-<= and noticed that * is not imported, so I added an import for that (and another useful property that was defined in an earlier chapter).

Do you guys think these imports should be added at the top of the chapter?  I can PR the change, if so.

negatratoron:
Closes #1079

### DIFF
--- a/src/plfa/part1/Relations.lagda.md
+++ b/src/plfa/part1/Relations.lagda.md
@@ -15,8 +15,8 @@ the next step is to define relations, such as _less than or equal_.
 ```agda
 import Relation.Binary.PropositionalEquality as Eq
 open Eq using (_≡_; refl; cong)
-open import Data.Nat using (ℕ; zero; suc; _+_)
-open import Data.Nat.Properties using (+-comm; +-identityʳ)
+open import Data.Nat using (ℕ; zero; suc; _+_; _*_)
+open import Data.Nat.Properties using (+-comm; +-identityʳ; *-comm)
 ```
 
 


wadler (changes requested):
Thanks.

*-comm should not be imported, not least because it is one of the later exercises.

negatratoron:
Isn't *-comm an earlier exercise, in Induction?

I had added it because I used it in the proof, and figured it was reasonable to have that in the imports since +-comm is used in the example proof of +-mono-<=.

I did remove it from the PR, but IMO it should be present, even though it serves as a hint if the reader looks closely at the imports.

negatratoron:
While I have you here, thanks for all your effort on this book by the way!  It's really well-written, and I also appreciate all the talks you've given that I can watch on youtube.

negatratoron:
@wadler Just a reminder, I made the change and I'm perfectly happy with it, go ahead and merge if you're ready

wadler (changes requested):
Thanks for the updates, as well as the kind words.

You are correct, *-comm is proved in Induction not Relations, and since +-comm is used to prove +-mono it's natural to use *-comm to prove *-mono. Unless you know a really nice proof that doesn't use *-comm (I don't), best to restore *-comm to the imports. Apologies for the flip flop, and thanks for the correction.

negatratoron:
No problem, pushed again, thanks!

wadler (approved):
Thank you!

PR #1081, opened 2025-01-11
Switch trans' from quote to prime in text to match agda code

negatratoron:
Just another tiny edit.  Reading the text I searched for `trans'` to refer to it, but got no results and thought at first the definition was missing.  Actually, the definition uses the prime character whereas the name in the text used a single quote.

This makes the name in the agda code reflexively equal to the name in the text :)


wadler (approved):
Good catch! Thank you.

PR #1087, opened 2025-01-24
Fix typo in `⊢plus`

ivanovs-4:


wadler (changes requested):
Thank you for the attempt to help, but I believe the original is correct and your fix is incorrect. If you try to execute it in Agda, you will see that it doesn't type check. This is one reason why I try to minimise places in PLFA that Agda is quoted without actually being type-checked.

ivanovs-4:
Indeed, I executed the code in Agda and looked more carefully at the definition of ⊢case and ⊢zero constructors.
Now I see why my fix is incorrect.

Thank you for this wonderful learning material!

wadler:
Thank you for the kind words. I'm sure I'll see other pull requests from you in future that we'll be able to accept.

Issue #1088, opened 2025-02-05
Some difficulties in the first part, as a beginner


cjauvin:
I am a total beginner in Agda, slowly going through the first section on Naturals. Because I'm an experienced Emacs user, I'm doing all the exercices interactively, but when I reached that part, I was a bit confused:

https://github.com/plfa/plfa.github.io/blob/1824aed8705818759127508de513918edb1c60bb/src/plfa/part1/Naturals.lagda.md?plain=1#L789

Just before it says to type it, so this is actually the first time that I realized that the code must be properly located inside an \`\`\`agda .. \`\`\` block (at least in the context of a Markdown file, I gather), and when I did that and evaluated it, I got an error saying that `_+_` was already defined elsewhere in the file. Then I changed it to `_$_`, and it worked. Not sure what is the exact learning intention, or the proper way to make this more clear, but I thought I would mention it, in the hope that it can be helpful.

wenkokke:
The passage in question was written to explain interactive editing, but perhaps not to be interactively followed along with.

Perhaps we could clarify that, if the user wishes to type along, they should create a fresh Agda file, but that brings with it a host of other bureaucracy.

It may be better to simply clarify that this is intended to illustrate interactive editing, but shouldn't be followed.

cjauvin:
It's not a super big deal, but yes, I think it would help to clarify a bit, especially given that the file, previous to that point, is meant to be editable ("insert your code here") and interactive, and it works quite well.

wadler:
@cjauvin 

Thanks for the feedback. Good point. The paragraph might begin along these lines: "Let's consider how to define addition interactively. (If you want to follow along, this replaces the definition above.)" And then continue as before. Can you  please register a pull request along these lines? Please edit to suit to avoid your previous confusion.

cjauvin:
@wadler Yes absolutely, I will do it with pleasure, I'm very happy to be able to contribute to this fine and very interesting work.

cjauvin:
Just in case you weren't notified of it for some reason: https://github.com/plfa/plfa.github.io/pull/1089

PR #1089, opened 2025-02-05
Improve the interactivity of the section on holes in Naturals

cjauvin:
As discussed in https://github.com/plfa/plfa.github.io/issues/1088, I suggest this small change to improve the interactivity of the file on Naturals a bit, and hopefully remove the confusion that I at least had with it.

As I explain in the text, the only price to pay is that we have to use `_+'_` (plus prime) instead of `_+_`, because it was already defined in the file.

### DIFF
--- a/src/plfa/part1/Naturals.lagda.md
+++ b/src/plfa/part1/Naturals.lagda.md
@@ -785,12 +785,12 @@ in combination provide features that help to create definitions
 and proofs interactively.
 
 To demonstrate it, we will use addition again, but since `_+_` was already
-defined in the file (above), we will use `_+'_` (plus prime) this time. You
+defined in the file (above), we will use `_plus_` this time. You
 can first uncomment the following code (remove the `--`s):
 
 ```agda
--- _+'_ : ℕ → ℕ → ℕ
--- m +' n = ?
+-- _plus_ : ℕ → ℕ → ℕ
+-- m plus n = ?
 ```
 
 The question mark indicates that you would like Agda to help with
@@ -798,8 +798,8 @@ filling in that part of the code. Once uncommented, if you type `C-c C-l`
 (pressing the control key while hitting the `c` key followed by the `l`
 key), which stands for **l**oad, the question mark will bereplaced:
 
-    _+'_ : ℕ → ℕ → ℕ
-    m +' n = { }0
+    _plus_ : ℕ → ℕ → ℕ
+    m plus n = { }0
 
 The empty braces are called a *hole*, and 0 is a number used for
 referring to the hole.  The hole will display highlighted in green.
@@ -819,9 +819,9 @@ You will be given the prompt:
 Typing `m` will cause a split on that variable, resulting
 in an update to the code:
 
-    _+'_ : ℕ → ℕ → ℕ
-    zero +' n = { }0
-    suc m +' n = { }1
+    _plus_ : ℕ → ℕ → ℕ
+    zero plus n = { }0
+    suc m plus n = { }1
 
 There are now two holes, and the window at the bottom tells you the
 required type of each:
@@ -839,9 +839,9 @@ required type of the hole, and what free variables are available:
 This strongly suggests filling the hole with `n`.  After the hole is
 filled, you can type `C-c C-space`, which will remove the hole:
 
-    _+'_ : ℕ → ℕ → ℕ
-    zero +' n = n
-    suc m +' n = { }1
+    _plus_ : ℕ → ℕ → ℕ
+    zero plus n = n
+    suc m plus n = { }1
 
 Again, going into hole 1 and typing `C-c C-,` will display information on the
 required type of the hole, and what free variables are available:
@@ -859,9 +859,9 @@ you might use, if there is a choice.  In this case, it displays the following:
 
 Filling the hole with `suc ?` and typing `C-c C-space` results in the following:
 
-    _+'_ : ℕ → ℕ → ℕ
-    zero +' n = n
-    suc m +' n = suc { }1
+    _plus_ : ℕ → ℕ → ℕ
+    zero plus n = n
+    suc m plus n = suc { }1
 
 Going into the new hole and typing `C-c C-,` gives similar information to before:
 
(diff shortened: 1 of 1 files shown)


wenkokke:
We tend to use `’` (typed using `\'`) for primes. However, decorating the `+` with a prime would require us to explain that before we even explain interactive editing. Therefore, I'd suggest that we don't alter the explanation in this manner, but rather instruct the reader to comment out the above definition.

cjauvin:
@wadler, in the issue thread, asked me to propose a way to reduce my own confusion, which is what I tried to do with this PR:

1. Up to that point in that document (the object of this PR), I was using the code in the agda blocks, including the exercices, etc.
2. At that particular point, it is said to "type some code", but there wasn't an agda block for it, so I added it
3. The code in this block must be commented out at first, because the reader must first read about holes and "?", before evaluating it and seeing what happens
4. However, once that is done, there remains the problem that `_+_` is already defined in the document, so I decided to go with `_+'_`, but of course it could be anything else

cjauvin:
> but rather instruct the reader to comment out the above definition.

Oh sorry, I understood what you meant just after posting my previous comment! You're right, it's a possibility also, although I'm not sure it would be easier, especially if you did, like me, all the exercices already? Maybe another operator than `_+'_` would be more acceptable to you? I considered `_add_`, but thought it was a bit verbose.

wadler:
Please re-read my original comment and tell me what you think was wrong with the wording I proposed there.

cjauvin:
You suggested this:

> Let's consider how to define addition interactively. (If you want to follow along, this replaces the definition above.)

It's probably due to my lack of understanding of how Agda works, but I don't see how it's possible to define addition once more (replacing it), since it was previously defined there:

https://github.com/plfa/plfa.github.io/blob/1824aed8705818759127508de513918edb1c60bb/src/plfa/part1/Naturals.lagda.md?plain=1#L306C8-L309

and we are assuming that the user had already evaluated the whole file (using `C-c C-l`), to complete the previous exercices. I checked if there is a mechanism to un-define a symbol, like some languages have, but didn't find any. So given these constraints, I don't really understand what you have in mind.

wenkokke:
> You suggested this:
> 
> 
> 
> > Let's consider how to define addition interactively. (If you want to follow along, this replaces the definition above.)
> 
> 
> 
> It's probably due to my lack of understanding of how Agda works, but I don't see how it's possible to define addition once more (replacing it), since it was previously defined there.

I believe @wadler's comment is meant to suggest that the student simply edits the previous definition.

cjauvin:
So I tried to do as you suggest: go back and edit the previous definition of `_+_` at:

https://github.com/plfa/plfa.github.io/blob/1824aed8705818759127508de513918edb1c60bb/src/plfa/part1/Naturals.lagda.md?plain=1#L307-L309

that is, I commented out the previous definition and replaced it with the one with "?". The moment I do that and try to evaluate it (`C-c C-l`), I run into other problems, in particular with 

https://github.com/plfa/plfa.github.io/blob/dev/src/plfa/part1/Naturals.lagda.md?plain=1#L885

So unfortunately I must give up with this PR. It's quite likely that there's something I don't understand, but from my current limited perspective, what I've been proposing is the only thing that works.

wenkokke:
> So I tried to do as you suggest: go back and edit the previous definition of `_+_` at:
> 
> https://github.com/plfa/plfa.github.io/blob/1824aed8705818759127508de513918edb1c60bb/src/plfa/part1/Naturals.lagda.md?plain=1#L307-L309
> 
> that is, I commented out the previous definition and replaced it with the one with "?". The moment I do that and try to evaluate it (`C-c C-l`), I run into other problems, in particular with
> 
> https://github.com/plfa/plfa.github.io/blob/dev/src/plfa/part1/Naturals.lagda.md?plain=1#L885
> 
> So unfortunately I must give up with this PR. It's quite likely that there's something I don't understand, but from my current limited perspective, what I've been proposing is the only thing that works.

Hm, that makes sense. Agda does some magic internally to make computing with naturals (as defined in this chapter) actually fast, for which is does a lot of checks to make sure that `+` actually is addition.

The best solution might just be to keep the writing as-is, but tell the student that if they want to follow along, they should call their `+` something else, like `plus`, to avoid naming conflicts.

cjauvin:
> The best solution might just be to keep the writing as-is, but tell the student that if they want to follow along, they should call their + something else, like plus, to avoid naming conflicts.

As a last attempt, I've replaced my initial `_+'_`, which you didn't like, with `_plus_`. Me personally, I find that it's slightly clearer. overall with this change, but I agree that it's quite a minor point in the end, so maybe it's not worth to change.

wenkokke:
> > The best solution might just be to keep the writing as-is, but tell the student that if they want to follow along, they should call their + something else, like plus, to avoid naming conflicts.
> 
> 
> 
> As a last attempt, I've replaced my initial `_+'_`, which you didn't like, with `_plus_`. Me personally, I find that it's slightly clearer. overall with this change, but I agree that it's quite a minor point in the end, so maybe it's not worth to change.

My main point was that the text describing the various stages of interactive editing (i.e. all the + symbols) should be left the same—because the case we're optimising for is reading but _not_ interactively following along—and that the suggestion to use a different name if one is keen to follow along should only appear once in the text itself.

Hence, something along the lines of:

> Let's consider how to define addition interactively. (If you want to follow along, use a name other than `_+_` to avoid conflict with the definition above.)

cjauvin:
So I reverted my changes and used your suggestion. I understand your point, but in the end I still respectfully disagree: I think a lot of experienced programmers will be confused if you don't make a clear distinction between runnable code blocks and blocks that look like code, but are only meant as explanation. Another example of this is here:

https://github.com/plfa/plfa.github.io/blob/1824aed8705818759127508de513918edb1c60bb/src/plfa/part1/Naturals.lagda.md?plain=1#L90-L93

When I first read that, I wondered: could that horizontal line be part of the syntax of this language? Is that code, or something else? From the context at this point, it's not entirely clear.

Anyway I really appreciate that you were willing to have a discussion about fine details like that here, and I will continue reading your book with interest!

wenkokke:
> I think a lot of experienced programmers will be confused if you don't make a clear distinction between runnable code blocks and blocks that look like code, but are only meant as explanation.

There are two ways to distinguish these in the current book. When editing interactively, code blocks are demarcated by

``````markdown
```agda
-- code goes here
```
``````

whereas specification blocks are indented by four spaces:

``````markdown
    specification goes here
``````

Secondly, when reading the book, code blocks benefit from syntax highlighting

Screenshot 2025-02-08 at 17 24 16

whereas specification blocks do not:

Screenshot 2025-02-08 at 17 24 01

@wadler It would perhaps be useful to mention this explicitly.

wadler (dismissed):
This seems fine. Thanks to @cjauvin for determining that replacing the previous definition won't work, and to @wenkokke for the suggested phrasing (and everything else she does). I think I'd add a phrase: "use a name other than _+_, such as _+′_, ..."

wadler:
"@wadler It would perhaps be useful to mention this explicitly." Good idea. Would one of you like to write a pull request?

wenkokke (approved):

PR #1096, opened 2025-02-11
Fix code formatting for an exercise in Equality

fangyi-zhou:


wadler (approved):
Good catch. Thank you!

PR #1098, opened 2025-02-12
Fix a sentence regarding pattern matching in Connectives

### LINKED ISSUE #1097: Contradicting paragraph in Connectives
In the section of Truth

> As with the product, the record type `⊤` and the data type `⊤′` behave similarly, but η-equality holds by definition for the record type. While proving `η-⊤′`, we do not have to pattern match on `w`---Agda knows it is equal to tt′:
> 
> ```agda
> η-⊤′ : ∀ (w : ⊤′) → tt′ ≡ w
> η-⊤′ tt′ = refl
> ```
> The pattern matching on the left-hand side is essential. Replacing `w` by `tt′` allows both sides of the propositional equality to simplify to the same term. As with products, it is convenient to have η-equality definitionally, so we use records in preference to data types whenever there is only one constructor.

It seems that the sentence before the code block suggests the pattern match on `w` is not necessary, but the sentence afterwards suggests the pattern match is essential. I believe the sentence preceeding the code block is meant for the record type, and should be moved to earlier part of the text (removing the primes as necessary).

Could you confirm my understanding is correct? If so, happy to raise a pull request to fix.

fangyi-zhou:
Closes #1097

### DIFF
--- a/src/plfa/part1/Connectives.lagda.md
+++ b/src/plfa/part1/Connectives.lagda.md
@@ -258,8 +258,9 @@ Given evidence that `⊤` holds, there is nothing more of interest we
 can conclude.  Since truth always holds, knowing that it holds tells
 us nothing new.
 
-The nullary case of `η-×` is `η-⊤`, which asserts that any
-value of type `⊤` must be equal to `tt`:
+The nullary case of `η-×` is `η-⊤`.
+While proving `η-⊤`, we do not have to pattern match on `w`---Agda *knows* it
+is equal to `tt`:
 ```agda
 η-⊤ : ∀ (w : ⊤) → tt ≡ w
 η-⊤ w = refl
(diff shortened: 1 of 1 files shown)


wadler (dismissed):
The reworded version is clearly better. Thank you!

wenkokke:
I've pushed a small addition to your branch, since I felt that the text ended at an odd place, saying eta equality holds definitionally for records right before giving the data type definition. I've rewritten the text to:

``````markdown
As with the product, the record type `⊤` and the data type `⊤′` behave
similarly, but while η-equality holds *by definition* for the record type,
it does not for the data type, so we need to pattern match on `w`:
```agda
η-⊤′ : ∀ (w : ⊤′) → tt′ ≡ w
η-⊤′ tt′ = refl
```
``````

@fangyi-zhou @wadler Thoughts?

src/plfa/part1/Connectives.lagda.md
fangyi-zhou, line 1: Not in the scope of this change, but L134 uses the phrase "pattern matching" as a noun instead of "pattern match". Better to keep the consistency with changed line of L291

wenkokke: Removed that change, since you're right that I'll have to go in and make that change more consistently.

wadler: I like Wen's proposed rewrite. I think using "pattern match" here is fine. It is a verb rather than a noun phrase.

Issue #1099, opened 2025-02-12
eta equality should be motivated/explained


fangyi-zhou:
In #1051 and #1052 (Presenting conjunction, truth and existentials using records first before using data types), the text mentions eta equality without much motivation or explanation. I suggest that the notion of eta equality to be introduced/mentioned either in the Connectives chapter or (maybe more appropriately) the Equality chapter.

wenkokke:
That makes a lot of sense.

@wadler Thoughts?

wadler:
The suggestion is that eta-equality be introduced in Connectives, but that is already where it is introduced.

I think Connectives is a better choice than Equality, because it is hard to talk about the eta-rule for products until one talks about products.

I'd be happy to consider a pull request with suggested text to improve the motivation.

wenkokke:
> The suggestion is that eta-equality be introduced in Connectives, but that is already where it is introduced.

![Image](https://github.com/user-attachments/assets/41e46654-9129-4879-8915-6d0201b99284)

I think the issue is that it isn't really explained. There's the matter of introduction and elimination rules, then there's a bit of Agda code that uses the letter η, and then the next line talks about η-equality.

Even a roughshod explanation of β- and η-equality rules would help, such as saying that there's two kinds of equality rules on objects formed by introduction and elimination rules: β-equality rules that say "elim (intro x) = x" and η-equality rules that say "intro (elim x) = x" (except phrased with a bit more care and less pseudo-formality).

fangyi-zhou:
I've put a simple sentence for eta-equality in #1102, please let me know how it sounds!

P.S. Is there a canonical reference for the phrase "eta-equality" in the general sense? I learnt this concept in lambda calculus as (\x -> f x) == f, but I didn't know it generalises to introduction and elimination rules in general.

PR #1102, opened 2025-02-19
Add a sentence to motivate eta-equality

fangyi-zhou:
Addresses #1099.

### DIFF
--- a/src/plfa/part1/Connectives.lagda.md
+++ b/src/plfa/part1/Connectives.lagda.md
@@ -84,10 +84,15 @@ holds---how to _use_ the connective.[^from-wadler-2015]
 
 Applying each destructor and reassembling the results with the
 constructor is the identity over products:
+
 ```agda
 η-× : ∀ {A B : Set} (w : A × B) → ⟨ proj₁ w , proj₂ w ⟩ ≡ w
 η-× w = refl
 ```
+
+In the general sense, for a record, apply projections to all fields, and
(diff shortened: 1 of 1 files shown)


wadler (changes requested):
Thank you!

In "put them back" grammatically "them" refers to either "the destructors" or "the constructors", neither of which I think you intend. How about replacing the first sentence with this:

Start with a product, apply the destructors to it, and then apply the constructor; the result is the original product.

fangyi-zhou:
Thanks for the suggestion, my intention was the generalise the concept from the previous sentence to motivate the eta-quality law, since the preceeding sentence reads:

> Applying each destructor and reassembling the results with the
constructor is the identity over products:

In the spirit of providing more clarity, what about this:

In the general sense, with a record, apply the projections to all fields, and then apply the record constructor; the result is the original record.

wenkokke:
> Thanks for the suggestion, my intention was the generalise the concept from the previous sentence to motivate the eta-quality law, since the preceeding sentence reads:
> 
> > Applying each destructor and reassembling the results with the
> > constructor is the identity over products:
> 
> In the spirit of providing more clarity, what about this:
> 
> In the general sense, with a record, apply the projections to all fields, and then apply the record constructor; the result is the original record.

"For any record, projecting out each field and reassembling the results using the constructor should be the identity."

Maybe followed by (except using more appropriate unicode):

```
c (p1 r) ... (pN r) = r
```

PR #1111, opened 2025-04-30
Update TableOfContents.md

proglang:
Course @proglang in 2025


wenkokke:
There's something delightful about submitting this PR as the @proglang group.

PR #1130, opened 2025-09-14
Generalize `commute-subst-rename`

damhiya:
Hello! I generalized the statement of `commute-subst-rename` and added some definitions and lemmas in the Substitution appendix.

Here is the previous statement of `commute-subst-rename`:
```agda
commute-subst-rename : ∀{Γ Δ}{M : Γ ⊢ ★}{σ : Subst Γ Δ}
                        {ρ₁ : Rename Γ (Γ , ★)}{ρ₂ : Rename Δ (Δ , ★)}
                     → (∀{x : Γ ∋ ★} → exts σ (ρ₁ x) ≡ rename ρ₂ (σ x))
                     → subst (exts σ) (rename ρ₁ M) ≡ rename ρ₂ (subst σ M)
```

And suggested revision:
```agda
Commute : ∀ {Γ₁ Γ₂ Δ₁ Δ₂} → Subst Γ₂ Δ₂ → Rename Γ₁ Γ₂ → Subst Γ₁ Δ₁ → Rename Δ₁ Δ₂ → Set
Commute {Γ₁ = Γ₁} σ₁ ρ₁ σ₂ ρ₂ = ∀ (x : Γ₁ ∋ ★) → σ₁ (ρ₁ x) ≡ rename ρ₂ (σ₂ x)

commute-subst-rename : ∀ {Γ₁ Γ₂ Δ₁ Δ₂}
                         {σ₁ : Subst Γ₂ Δ₂} {ρ₁ : Rename Γ₁ Γ₂} {σ₂ : Subst Γ₁ Δ₁} {ρ₂ : Rename Δ₁ Δ₂}
                     → Commute σ₁ ρ₁ σ₂ ρ₂
                     → ∀ (M : Γ₁ ⊢ ★) → subst σ₁ (rename ρ₁ M) ≡ rename ρ₂ (subst σ₂ M)
```

Old statement is an instance of the new one, where `σ₁ = exts σ` and `σ₂ = σ`.

The motivation for this change is two folded. First, it is simply more general, so it has more applications in theory. Second and more importantly, the new proof can be naturally adapted to typed setting, while the older one is not. Consider the previous proof of `commute-subst-rename` for the `ƛ_` case.
```agda
commute-subst-rename {Γ}{Δ}{ƛ N}{σ}{ρ₁}{ρ₂} H =
   cong ƛ_ (commute-subst-rename{Γ , ★}{Δ , ★}{N}
               {exts σ}{ext ρ₁}{ext ρ₂} (λ {x} → H′ {x}))
   where
   H′ : {x : Γ , ★ ∋ ★} → exts (exts σ) (ext ρ₁ x) ≡ rename (ext ρ₂) (exts σ x)
   H′ = ...
```
In the type of `H′`, `exts σ` on the LHS and `exts σ` on the RHS may seem equal, but each occurrence of `exts` operations are responsible for different variables. If we annotate `ext/exts` operations with relatedness, the type would be something like the following:
```agda
H′ : {x : Γ , ★ ∋ ★} → exts₂ (exts₁ σ) (ext₂ ρ₁ x) ≡ rename (ext₂ ρ₂) (exts₂ σ x)
```
However the use of `H′` in `commute-subst-rename{Γ , ★}{Δ , ★}{N}{exts σ}{ext ρ₁}{ext ρ₂} (λ {x} → H′ {x})` requires `exts₁ σ` and `exts₂ σ` to be judgementally equal. This is the case in our untyped setting, but it could be false in general, for instance in intrinsically typed setting. The generalized version does not have this problem.

I also fixed text according to changed proof structure.

### DIFF
--- a/src/plfa/part2/Substitution.lagda.md
+++ b/src/plfa/part2/Substitution.lagda.md
@@ -669,70 +669,72 @@ compose-rename {Γ}{Δ}{Σ}{A}{ƛ N}{ρ}{ρ′} = cong ƛ_ G
 compose-rename {M = L · M} = cong₂ _·_ compose-rename compose-rename
 ```
 
-The next lemma states that if a renaming and substitution commute on
-variables, then they also commute on terms. We explain the proof in
-detail below.
+We define a commutation relation `Commute σ₁ ρ₁ σ₂ ρ₂` between pairs `(σ₁, ρ₁)` and `(σ₂, ρ₂)`,
+states that for every variable `x`, applying `ρ₁` then `σ₁` is the same as applying `σ₂` then `ρ₂`.
 
+```agda
+Commute : ∀ {Γ₁ Γ₂ Δ₁ Δ₂} → Subst Γ₂ Δ₂ → Rename Γ₁ Γ₂ → Subst Γ₁ Δ₁ → Rename Δ₁ Δ₂ → Set
+Commute {Γ₁ = Γ₁} σ₁ ρ₁ σ₂ ρ₂ = ∀ (x : Γ₁ ∋ ★) → σ₁ (ρ₁ x) ≡ rename ρ₂ (σ₂ x)
+```
+
+An example of the `Commute` relation is `σ₁ = exts σ`, `ρ₁ = S_`, `σ₂ = σ`, and `ρ₂ = S_`,
+where `σ` is arbitrary substitution.
+
+```agda
+Commute-S : ∀ {Γ Δ} (σ : Subst Γ Δ)
+          → Commute {Γ₁ = Γ} {Γ₂ = Γ , ★} {Δ₁ = Δ} {Δ₂ = Δ , ★} (exts σ) S_ σ S_
+Commute-S σ x = refl
+```
+
+The `Commute` relation is closed under `ext/exts` operation.
+We prove it by case analysis on the input variable:
+
+```agda
+Commute-ext : ∀ {Γ₁ Γ₂ Δ₁ Δ₂}
+                {σ₁ : Subst Γ₂ Δ₂} {ρ₁ : Rename Γ₁ Γ₂} {σ₂ : Subst Γ₁ Δ₁} {ρ₂ : Rename Δ₁ Δ₂}
+            → Commute {Γ₁ = Γ₁} σ₁ ρ₁ σ₂ ρ₂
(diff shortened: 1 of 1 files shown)


wadler:
Thanks, @damhiya . I'll leave @jsiek to comment. (Also, there are some failing checks @damhiya  might want to look at.)

damhiya:
Thank you for the quick reply!

I read the build error message, but it seems like that the error is irrelevant to change made in this PR. (https://github.com/plfa/plfa.github.io/pull/1129 also fails check with same error)
```
Error: cabal: Failed to build shoggoth-0.1.0.0 (which is required by
exe:builder from plfa-22.8). See the build log above for details.
```

jsiek:
Thank you @damhiya for this PR. I like your generalization.
I have one suggestion, which is to keep the old theorem as a corollary of the new generalized theorem.
Then the uses of the old theorem do not need to change and don't need to use the Commute-S lemma.

damhiya:
@jsiek Thank you, I recovered the original theorem as your suggestion!

src/plfa/part2/Substitution.lagda.md
```diff
@@ -767,7 +769,7 @@ The proof proceed by cases on `x`.
   and sequencing.
 
 * If `x = S x`, we unfold the first use of `exts` and sequencing, then
-  apply the lemma `commute-subst-rename`. We conclude by the
+  apply the lemma `Commute-subst-rename` and `Commute-S`. We conclude by the
```
jsiek: Please remove this change.

src/plfa/part2/Substitution.lagda.md
```diff
@@ -754,7 +756,7 @@ exts-seq = extensionality λ x → lemma {x = x}
        (exts σ₁ ⨟ exts σ₂) (S x)
      ≡⟨⟩
        ⟪ exts σ₂ ⟫ (rename S_ (σ₁ x))
-     ≡⟨ commute-subst-rename{M = σ₁ x}{σ = σ₂}{ρ₁ = S_}{ρ₂ = S_} refl ⟩
+     ≡⟨ Commute-subst-rename (Commute-S σ₂) (σ₁ x) ⟩
```
jsiek: Please remove this change.

jsiek (changes requested):
Please change the uses of the lemma back to the old one. Thank you!

jsiek (approved):
Looks great, thanks again!

PR #1137, opened 2025-10-24
Add tips page to the book 2

noughtmare:
I've rebased and addressed the comments of #881.


noughtmare:
There's a problem upstream: https://github.com/kazu-yamamoto/crypton-certificate/issues/21

Although I think we should drop these very old GHC versions. 9.6 is already old at this point.

wadler:
I'm not sure how I feel about this tips page. On the one hand, I'm sure some students will find it useful. On the other hand, it is quite low level and I'm not sure I agree with all of the advice. Wen and Jeremy, what do you think?

wadler:
(Also, thanks to Michael and Noughtmare for putting this together!)

PR #1192, opened 2026-05-02
Fix import mistake in Decidable: import ≤ instead of <

attoknot:
Hi

This feels like a typo since those constants are never used on that page except in an exercise.


attoknot:
Nevermind. The type `_≤_`. is redefined in the next paragraph. For some reason my brain totally ignored it.

Issue #1203, opened 2026-06-03
Simpler proof of Isomorphism.≲-antisym


peterthiemann:
I was surprised to discover that the proof of Isomorphism.≲-antisym can be simplified.
Here is the original proof:

https://github.com/plfa/plfa.github.io/blob/8e5406ee801299ff29af575277f2f4d4bea157e0/src/plfa/part1/Isomorphism.lagda.md?plain=1#L397-L421

Interestingly, one can pattern match against the arguments `to≡from` and `from≡to` resulting in this shorter proof:

```
≲-antisym A≲B B≲A to≡from@refl from≡to@refl =
  record
    { to      = to A≲B
    ; from    = from A≲B
    ; from∘to = from∘to A≲B
    ; to∘from = from∘to B≲A
    }
```

Matching against `refl` works only because records are subject to eta-expansion.
What is the argument for having the longer proof in the text?

wadler:
Neatly observed! Like you, I'm surprised this works. Please put in a pull request to update the proof. The pull request should include text to add to PLFA to explain to its readers the lesson they should take from the example. I think you are in a better place than me to extract the lesson, but one possible lesson is this: pattern match against refl when you can, and be aware of how eta expansion can make this possible when otherwise it might not work.

peterthiemann:
Do you want me to replace the current proof or discuss the direct matching version as an alternative?

PR #1204, opened 2026-06-04
Isomorphism antisym

peterthiemann:
Proposed new language to address issue https://github.com/plfa/plfa.github.io/issues/1203


wadler:
Thanks! Three comments.

1. Suggested edit: "A value A≲B is definitionally ..." --> "A record value is definitionally ..."

2. I didn't understand the explanation of how eta expansion of records enables the use of records. If one applies the eta expansion to `A≲B` and then simplifies `to A≲B` the result is `to A≲B`, so I fail to see how this helps.

3. I think it's overkill to define `≲-antisym` twice. What would you think about removing the first definition and replacing it by the second?

peterthiemann:
All good suggestions! Implemented.

